PR #1, opened 2019-03-31
suggestions for AEP0

ltalirz:
contains both fixes of typos and slight modifications

### DIFF
--- a/0_aep_guidelines/aep_guidelines.md
+++ b/0_aep_guidelines/aep_guidelines.md
@@ -2,67 +2,86 @@
 
 | AEP number | 0                                                            |
 |------------|--------------------------------------------------------------|
-| Title      | AiiDA Enhancement Proposal Guidelines                        |
-| Authors    | [Leopold Talirz](mailto:leopold.talirz@epfl.ch),  [Kevin M. Jablonka](mailto:kevin.jablonka@epfl.ch) |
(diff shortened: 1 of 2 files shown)


0_aep_guidelines/aep_guidelines.md
```diff
@@ -2,67 +2,86 @@
...
+A summary for pros and cons for the proposed enhancement, detailing e.g. which possible compatibility issues may arise.
+
+### Submission process
+1. [Fork](https://help.github.com/en/articles/fork-a-repo) this repository
+2. Create a folder with the title of the AEP in lower [snake-case](https://en.wikipedia.org/wiki/Snake_case)
+   and put a [Markdown](https://guides.github.com/features/mastering-markdown/) file with the title of the AEP in lower case inside this folder.
+     * You may use the `aep_guidelines.md` as a template.
+     * You may want to include additional files relevant to your enhancement proposal.
+3. Submit a [pull request](https://help.github.com/en/articles/creating-a-pull-request) to the AEP repository
+4. If the pull request is accepted the status will change to `active` and it will be added to the AEP index in the `README.md`
```
ltalirz: This is different from the jupyter project, right?
I see a lot of "submitted" PRs in their repo: https://github.com/jupyter/enhancement-proposals
Also, how does this work - either the PR is made with "submitted" or it's made with "active", right?

ltalirz: I guess the PR should already add it to the README.md, and it should probably use the "submitted" status.

We could have as a convention (not necessarily a rule) that the next PR (from someone else) that changes that AEP should also change the status to "active" - at this point at least two people will have been involved.

PR #5, opened 2019-09-10
use readme.md as default filename

ltalirz:
this saves unnecessary renaming + will make github show the AEP
when you enter the folder.


kjappelbaum:
Good idea! Thanks!

PR #7, opened 2019-10-18
AEP: Abstract and improve the node file repository

sphuber:
 - [x] Status is `submitted`


001_file_repository_improvement/readme.md
```diff
@@ -0,0 +1,210 @@
...
+### Store index of node repository contents in database
+When storing a file in the repository for a node, instead of the node directly passing this to the repository interface, it should go through a node-repository manager.
+This class is the gateway between the node and its repository.
+When storing an object for a node in the repository, it will first request the configured repository to store the object with the given name.
+When successful, the repository will return a key that is the unique identifier for the stored object.
+This key is then used to create an entry in the node repository index table to register its existence.
+
+### Fine controlled API to bundle and compress multiple objects of a node's repository
+Each node has a repository which can contain many files with arbitrary hierarchy.
+Often these files do not have to be accessed very often and so it may be more beneficial to bundle and compress the files.
```
ltalirz: We should also mention that particularly when files are stored remotely, the cost of transfer will likely be comparable or longer than unzipping.

001_file_repository_improvement/readme.md
```diff
@@ -0,0 +1,210 @@
...
+        :param key: fully qualified identifier for the object within the repository
+        """
+
+
+class NodeRepositoryManager(object):
+    """Gateway between a `Node` instance and a `Repository`
+
+```
+Both the `Node` class as well as the `NodeRepositoryManager` should provide the same interface as the `Repository` class.
+Each `Node` instance will then instantiate a `NodeRepositoryManager` instance in its constructor and all repository calls will pipe through to the manager.
```
ltalirz: One manager instance per node instance?
That's adding significant overhead when you load many nodes. 
We should think whether this is necessary.

001_file_repository_improvement/readme.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Cons
+* Storing an index of all objects in a node's repository has a non-negligible overhead in database load as well as storage space.
+* The implementation of storing and retrieving objects from the repository becomes more complicated with the addition of bundling and compressing.
+* The storage of files in the repository becomes even more opaque for the user.
+  In the case of an object store this is necessary and not even a downside, but for users only using a local file system, this may be unintuitive.
+  However, even in the current situation the user is not supposed to directly access or interact with the repository on the local file system.
+
+## Open questions
+* Should the full configuration of a `Repository` be stored in its database entry, or the database just keeps a reference of known repositories.
```
ltalirz: That's a good question, even in a more general sense:
Which information should be stored in the config file, and which information should be stored in the database?

One clear rule is that any information that may be needed/useful without loading the database needs to be stored in the config file. E.g. if the information on the repo is stored in the DB only, then `verdi status`  can report the status of the repo only when it can also connect to the DB.

Another point to keep in mind is that repo connection info may contain sensitive credentials. Not sure whether this would also be pro-config, but at least we already have another case there (the DB connection info) the we'll want to solve at some point.

001_file_repository_improvement/readme.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Store index of node repository contents in database
+When storing a file in the repository for a node, instead of the node directly passing this to the repository interface, it should go through a node-repository manager.
+This class is the gateway between the node and its repository.
+When storing an object for a node in the repository, it will first request the configured repository to store the object with the given name.
+When successful, the repository will return a key that is the unique identifier for the stored object.
+This key is then used to create an entry in the node repository index table to register its existence.
+
+### Fine controlled API to bundle and compress multiple objects of a node's repository
+Each node has a repository which can contain many files with arbitrary hierarchy.
```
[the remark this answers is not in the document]
sphuber: No here I just mean to say that each node has a "repository" (which is a sub folder in the repository) where it stores content as files, just like it can store data in its attributes or extras etc.

001_file_repository_improvement/readme.md
sphuber: It would of course be lazily loaded when calling any of the repository methods on the `Node`. I do think that we should keep the interface that storing files in the repository for a node goes through the `Node` interface. Forcing users to load a different class and doing `repo.list_object_names(node)` seems like a bad idea. Other solution is to have a singleton node repository manager that is used by `Node` repo methods, but that also seems non-ideal. The `NodeRepositoryManager` class will be lightweight so cheap to instantiate and especially when lazily loaded I don't think it should pose a problem

001_file_repository_improvement/readme.md
sphuber: Meaning that the added cost of unzipping will be offset by the decrease in transfer time? So in terms of speed we don't lose anything but we reduce the required storage space?

001_file_repository_improvement/readme.md
sphuber: I agree about the information being available without having to load the database would probably be useful. I am also leaning towards putting it in the config file. We would have to look how this would work with export/import though. Most likely the details of a repository do not make sense for the person importing it as they might not have access to it.

Regarding sensitive data, I don't think that is a criterion and having that in the config file is not a problem. If you have access to the config file of an AiiDA installation, that means you have access to the machine and so it is already compromised. Because at the very minimum that config would always require database connection details, which means you would automatically have access to all repository configuration details in said database.

001_file_repository_improvement/readme.md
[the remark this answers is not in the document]
ltalirz: Ok, perhaps one could say something like

> Each node can have associated files with arbitrary directory structure.

and reserve the term `repository` to the profile-wide thing.

001_file_repository_improvement/readme.md
ltalirz: Yes, I just meant that, while you may "feel" the added time of unzipping with a local repository (we should do some tests, though), in the case of remote repositories, the added time may likely be inconsequential.

001_file_repository_improvement/readme.md
ltalirz: >  Other solution is to have a singleton node repository manager that is used by Node repo methods, but that also seems non-ideal. 

Could you elaborate on that? Will the `NodeRepositoryManager` have any state associated with a the `Node`? 
Anyhow, I think one should consider the singleton option - if it's a bad idea, fine to drop it.

001_file_repository_improvement/readme.md
ltalirz: > We would have to look how this would work with export/import though.

Could you elaborate on that? From my perspective, the existing way of exporting/importing via files could remain essentially unchanged, no?

Once we have support for an object store, we'll need to adapt the profile setup (`verdi setup`  and `verdi quicksetup`).

sphuber:
> Below a few comments and suggestions.

What is the typical procedure for an AEP PR? If I want to apply the changes I just make a new commit and push? Do we want to keep the commit history here? What is the typical approach?

ltalirz:
> @ltalirz what is the typical procedure for an AEP PR? If I want to apply the changes I just make a new commit and push?

Yes, that's the idea.

> Do we want to keep the commit history here? What is the typical approach?

I'd say let's keep the commit history at least in the PR. 
When merging, we can decide to squash it depending on the (e.g. have a look a the [commit history](https://github.com/jupyter/enhancement-proposals/commits/master) over at the jupyter folks).

Feel free to add me as a "champion" - then we should ask for reviews from all relevant parties, if you think you can even mention the `@aiidateam/plugin-developers`

For reference, here is how jupyter [propose to handle this review period](https://github.com/jupyter/enhancement-proposals/tree/master/29-jep-process#phase-2-rfc-for-the-jep)

001_file_repository_improvement/readme.md
sphuber: The import will become more complicated. Currently importing the repository consists merely of copying over the node folders to the repository on disk. In the future this will have to properly go through the `NodeRepositoryManager` such that the files are properly registered in the database. Also the `Repository` database entry with which they were associated in the original database will have no bearing on the person importing it, because it most likely won't have the same repository. Not necessarily a problem but just another difference to keep in mind.

001_file_repository_improvement/readme.md
sphuber: No, I think you are right that most likely it won't have to keep state so it might be feasible to have a singleton. Something like the `manager` instance which manages the database backend, It could maybe also manage the repository backend.

greschd:
I think this would be a good opportunity to also get rid of empty repository folders: The ``NodeRepositoryManager`` could be lazy in that it only creates a {folder, remote resource} when something is actually put in there.

The motivation for this is that FS performance (things like ``ls``, ``du``, ``find`` etc.) is impacted greatly by the number of folders in the repository, even if only ~1% of nodes actually contain data.

001_file_repository_improvement/readme.md
```diff
@@ -0,0 +1,211 @@
...
+    Attributes:
+    * name: Human readable label that also corresponds to the appropriate key in the config
+    * uuid: UUID associated with the repository upon creation
+
+    """
+    name = m.CharField(max_length=255, unique=True, blank=False)
+    uuid = UUIDField(auto=False, version=AIIDANODES_UUID_VERSION)
+
+
+class DbFile(m.Model):
```
chrisjsewell, line 71: for reference

chrisjsewell:
@sphuber, as discussed, perhaps this should be marked as rejected (at least for the exact spec you outlined here) and then merged/closed (?), and maybe you could add a note of why this was the case

001_file_repository_improvement/readme.md
sphuber: After the rewrite of the AEP to reflect the actual implementation, this no longer applies.

001_file_repository_improvement/readme.md
sphuber: After the rewrite of the AEP to reflect the actual implementation, this no longer applies.

sphuber:
@ltalirz @chrisjsewell @giovannipizzi I have updated the AEP to reflect the current implementation in `aiida-core`. I essentially rewrote the whole text, so all old comments no longer apply.

ltalirz (approved):
thanks for taking the time to wrap this up @sphuber  !

I haven't had the time to read through everything but since the implementation is already done I think it's great to have the thought process documented here.

Good to go from my side!

chrisjsewell:
just having a read...

007_improved_file_repository/readme.md
```diff
@@ -0,0 +1,350 @@
...
+The support of a (virtual) directory hierarchy, as provided by a local file system, is not assumed and therefore not required.
+The decoupling and abstraction of the file repository makes it possible, in theory, to implement it in the future for another backend than the `disk-objectstore`.
+
+However, this AEP intentionally does not go into detail how an alternative backend can be configured on a per-profile basis, but will assume that only support for the `disk-objectstore` will be provided for now.
+Nevertheless, the file repository is integrated in such a decoupled manner that it will be easy to implement this additional functionality in the future.
+Possible use-cases might be to have a file repository on a remote file system, accessed over SSH or even a traditional object store (like [S3](https://aws.amazon.com/s3/) or [SWIFT](https://wiki.openstack.org/wiki/Swift)) that communicate through REST APIs.
+
+## Detailed Explanation
+
+### Abstracting the file repository interface
```
chrisjsewell, line 55: I think it would be great here to have a basic before & after UML of this, similar (but not necessarily as detailed) to what I have done in https://github.com/aiidateam/aiida-core/pull/5172#issue-1021894297,
so people not already intimately knowledgable with AiiDA can get a quick idea of whats going on

chrisjsewell (approved):
few nitpicks

007_improved_file_repository/readme.md
sphuber: Would [this image](https://aiida.readthedocs.io/projects/aiida-core/en/develop/_images/schematic_design_class_hierarchy.png) I made for the documentation fulfill that need?

chrisjsewell: yep perfect 👍 Perhaps if you also had a "before" one as well

sphuber: Don't have a before, and there is not much to describe really. The code was more or less directly hard coded in the `Node` class.

README.md
```diff
@@ -1,25 +1,22 @@
...
+| 000    | active           | [AEP guidelines](000_aep_guidelines/readme.md)                                                         |
+| 001    | implemented      | [Drop support for Python 2.7](001_drop_python2/readme.md)                                              |
+| 002    | implemented      | [AiiDA Dependency Management](002_dependency_management/readme.md)                                     |
+| 003    | active           | [Adopt NEP 29](003_adopt_nep_29/readme.md)                                                             |
+| 004    | implemented      | [Infrastructure for importing completed `CalcJob`s](004_calcjob_importer/readme.md)                    |
+| 005    | draft            | [New Export Format](005_exportformat/readme.md)                                                        |
+| 006    | implemented      | [Efficient object store for the AiiDA repository](006_efficient_object_store_for_repository/readme.md) |
+| 007    | implemented      | [Abstract and improve the file repository](007_improved_file_repository/)                              |
```
chrisjsewell, line 19:
```suggestion
| 007    | implemented      | [Abstract and improve the file repository](007_improved_file_repository/readme.md)                              |
```

you reverted my fix

sphuber: Technically not a revert, since it is a new row. But I see the problem, just accidentally forgot.

sphuber: Fixed, guvnor

PR #8, opened 2019-11-15
Drop support for Python 2.7 with next major release

csadorf:
 - [x] Status is `submitted`

### DIFF
--- a/999_drop_python2_support/readme.md
+++ b/999_drop_python2_support/readme.md
@@ -25,7 +25,7 @@ We propose the following schedule:
  * Jan 1 2020: The support for Python 2.7 is officially discontinued.
    *Important* security and bug fixes are released on the 1.x line.
  * Mar 1 2020: The recommended moratorium on minor releases ends.
- * Jul 1 2020: Security and bug fix releases for the 1.x release line are  discontinued.
+ * Jul 1 2020: Security and bug fix releases for the AiiDA 1.x release line are  discontinued.
 
 The recommended moratorium on minor releases is supposed to reduce the potential for user confusion with respect to a change in feature set or potential backwards incompatibilities.
 However, the AiiDA team should revisit this policy with every new feature development and decide whether an immediate release of set feature outweighs the benefits of the moratorium.
--- a/999_drop_python2_support/readme.md
+++ b/999_drop_python2_support/readme.md
@@ -23,7 +23,7 @@ We propose the following schedule:
  * Dec 1 2020: The development of new features with support for Python 2.7 is discontinued.
    Any new features must be released as part of the AiiDA 2.x line.
  * Jan 1 2020: The support for Python 2.7 is officially discontinued.
-   *Important* security and bug fixes are released on the 1.x line.
+   *Important* security and bug fixes are released on the AiiDA 1.x line.
  * Mar 1 2020: The recommended moratorium on minor releases ends.
  * Jul 1 2020: Security and bug fix releases for the AiiDA 1.x release line are  discontinued.
 
--- a/999_drop_python2_support/readme.md
+++ b/999_drop_python2_support/readme.md
@@ -1,0 +1,99 @@
+# AiiDA Enhancement Proposal (AEP) Guidelines
+
+| AEP number | 000                                                          |
+|------------|--------------------------------------------------------------|
+| Title      | AiiDA Enhancement Proposal (AEP) Guidelines                  |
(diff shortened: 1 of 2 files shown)


csadorf:
I don't know how to add the type label to the PR.

999_drop_python2_support/readme.md
```diff
@@ -0,0 +1,69 @@
...
+The AiiDA team has stated previously that the support of AiiDA for Python 2.7 will dropped as of January 2020.
+
+This AEP lays out a concrete roadmap as to how this drop of support will be accomplished while aiming to find a reasonable balance between disruption of user workflows, developer burden, and public appearance of the project.
+
+## Proposed Enhancement
+
+We propose the following schedule:
+
+ * Dec 1 2020: The development of new features with support for Python 2.7 is discontinued.
+   Any new features must be released as part of the 2.x line.
```
ltalirz:
```suggestion
   Any new features must be released as part of the AiiDA 2.x line.
```

999_drop_python2_support/readme.md
```diff
@@ -0,0 +1,69 @@
...
+ * Jan 1 2020: The support for Python 2.7 is officially discontinued.
+   *Important* security and bug fixes are released on the 1.x line.
```
ltalirz:
```suggestion
   *Important* security and bug fixes are released on the AiiDA 1.x line.
```

999_drop_python2_support/readme.md
```diff
@@ -0,0 +1,69 @@
...
+ * Mar 1 2020: The recommended moratorium on minor releases ends.
+ * Jul 1 2020: Security and bug fix releases for the 1.x release line are  discontinued.
```
ltalirz:
```suggestion
 * Jul 1 2020: Security and bug fix releases for the AiiDA 1.x release line are  discontinued.
```

999_drop_python2_support/readme.md
```diff
@@ -0,0 +1,69 @@
...
+This has not only massively increased the pressure on users to update their tool chains, but also makes it excruciatingly more difficult to maintain support for Python 2.7 for AiiDA.
+
+There might be some expectation from users that a major version increase may imply a major change in feature set, however, we believe that the potential user confusion is far outweighed by the benefits of the expedited drop of support for Python 2.7.
+Furthermore, confusion can be reduced through transparent communication of the policy on the mailing list and as part of the change log.
```
ltalirz: Agree 100%.

ltalirz:
@sphuber @giovannipizzi Let us know your take as well!

999_drop_python2_support/readme.md
ltalirz: First thought this was a typo (since on is thinking about python 2 here)

giovannipizzi:
My gut feeling is
- if we want to accelerate dropping python 2, let's state (as you say) that we stop developing with python 2, we start dropping the six etc. dependencies already now in develop, and we aim for a fast release (on early Jan?) with only py3 support, but the smallest set of additional changes possible (i.e., we drop py2 in develop, and we branch off there for longer-term work, and merge after the release)
- then, I would be tempted to release a 1.1 release for this. I think releasing a 2.0 after 5 years of 0.x and 3 months of 1.x is a bit too much to digest and understand for users, that will not go and read the docs to understand why... Anyway, if we properly state which python versions we are compatible with, aren't *recent* versions of pip clever enough not to install the newer incompatible versions? For instance, I tried and if I do a `pip install -U scipy` in a py2 environment, I get 1.2.2 (while I get 1.3.2 in py3). So I don't feel there is a real need to go to a 2.0.

csadorf:
@giovannipizzi Sounds you would be more in favor of [this alternative AEP](https://github.com/aiidateam/AEP/pull/9)?

giovannipizzi:
> @giovannipizzi Sounds like you would be more in favor of this alternative AEP?

Correct sorry, I didn't realise there were two alternative AEPs!

Also, I'd suggest to protect the `master` branch with at least, say, 2 approvals, what do you think? (At the moment, I could merge)

sphuber:
> Correct sorry, I didn't realise there were two alternative AEPs!
> 
> Also, I'd suggest to protect the `master` branch with at least, say, 2 approvals, what do you think? (At the moment, I could merge)

I have added the branch protection to `master` requiring at least 2 approvals. Let's disucss this at the meeting this afternoon in any case

csadorf:
We'll move forward with #9 instead of this one.

PR #10, opened 2019-12-04
Dependency management

csadorf:
 - [x] Status is `submitted`


sphuber:
Thanks @csadorf few questions/comments:

* What is the difference between patch/hotfix? Is there just a difference in release protocol (i.e. the timeline) or also the used version numbers? If not the latter, can we just refer two both as patch releases?
* Should we include here some more detail in the exact policy of defining dependency requirements in `setup.json`. In the first commit of my [PR](https://github.com/aiidateam/aiida-core/pull/3597) I described what protocol I maintained. I will copy it here for ease:
```
The version requirements of all `install_requires` and `extras_require`
dependencies are reevaluated. Where we used to require explicit version
by pinning to exact versions using the `==` operator, we now use the more
liberal operator `~=`. This is the compatible release operator:

    https://www.python.org/dev/peps/pep-0440/#compatible-release

For a given release identifier `V.N` the compatible release clause is
approximately equivalent to the pair of comparison clauses:

    >= V.N, == V.*

With this rule we define our dependency requirements using the minor
version for packages with a major version of 1 or higher and using the
patch version for packages whose major version is still at 0. The reason
is that when packages have released `v1` they can be expected to respect
semantic versioning and not break backwards compatibility in minor
version whereas this is typically not the case for `v0` versions where
backward incompatible changes are introduced in minor versions as well.
Note that this approach will not prevent from breaking our builds
because packages break their API in minor versions, but at least we have
a consistent dependency requirement policy.
```
* Maybe a bit more detail on the introduction of the `requirements.txt` and the related testing. Does the CI against an installation from `requirements.txt` have to be done at every commit? Probably best if we can figure out how to do it only if it has changed?

ltalirz:
A suggested recipe for how to generate the `requirements.txt` would also be great.

ltalirz:
> Is the DM expected to also monitor for breaking changes in their dependencies? Is there a process for declaring which releases are still supported, and EOL'ing them [1]?

This is a good point - for 0.12.x we originally said 6 months after the release of 1.0, which puts this at 1-May-2020;
for 1.0.x we said 1-Jul-2020 in AEP 001.

csadorf:
> * What is the difference between patch/hotfix? Is there just a difference in release protocol (i.e. the timeline) or also the used version numbers? If not the latter, can we just refer two both as patch releases?

It depends on your [branch and release model](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow), but usually a hotfix *is* a patch-release that provides an important fix that needs to be urgently released and is therefore based directly on the latest tagged code version as opposed to a slightly more persistent branch that aggregates multiple fixes. The use of the term here indicates that such a fix should be **urgently** released (read: "as soon as possible"). 

It depends on the AiiDAs branching and release model whether there are other implications besides the urgency, but I would like to keep the distinction in the AEP.

> * Should we include here some more detail in the exact policy of defining dependency requirements in `setup.json`. In the first commit of my [PR](https://github.com/aiidateam/aiida-core/pull/3597) I described what protocol I maintained. I will copy it here for ease:
> 
> ```
> The version requirements of all `install_requires` and `extras_require`
> dependencies are reevaluated. Where we used to require explicit version
> by pinning to exact versions using the `==` operator, we now use the more
> liberal operator `~=`. This is the compatible release operator:
> 
>     https://www.python.org/dev/peps/pep-0440/#compatible-release
> 
> For a given release identifier `V.N` the compatible release clause is
> approximately equivalent to the pair of comparison clauses:
> 
>     >= V.N, == V.*
> 
> With this rule we define our dependency requirements using the minor
> version for packages with a major version of 1 or higher and using the
> patch version for packages whose major version is still at 0. The reason
> is that when packages have released `v1` they can be expected to respect
> semantic versioning and not break backwards compatibility in minor
> version whereas this is typically not the case for `v0` versions where
> backward incompatible changes are introduced in minor versions as well.
> Note that this approach will not prevent from breaking our builds
> because packages break their API in minor versions, but at least we have
> a consistent dependency requirement policy.
> ```

I would welcome a bit more detail as long as it is not too restrictive. Can you propose a concrete change or should I just add/adapt this paragraph?

> * Maybe a bit more detail on the introduction of the `requirements.txt` and the related testing. Does the CI against an installation from `requirements.txt` have to be done at every commit? Probably best if we can figure out how to do it only if it has changed?

I was under the impression that @ltalirz had some concrete ideas on how exactly to derive the `requirements.txt` file. @ltalirz Could you comment on this?

With respect to the CI pipeline. The language of the AEP already specifies that the "test pipeline for *releases branches*[emphasis added] is to be executed for both specifications". That implies that there is no need to execute both tests for *all* commits. 

I interpret this AEP as a slightly more general guideline that leaves some room for flexibility in implementation. I'm worried that we otherwise quickly run into a situation where we would need to regularly update it to keep up with changing setups, or alternatively have to ignore it. But there might be a different interpretation of course, and maybe we want to make this more concrete and update regularly? Would we need/want to seek community input for every update?

999_dependency_management/readme.md
```diff
@@ -0,0 +1,145 @@
+# AiiDA Enhancement Proposal (AEP) Guidelines
+
+| AEP number | 999                                                          |
+|------------|--------------------------------------------------------------|
+| Title      | AiiDA Enhancement Proposal (AEP) Guidelines                  |
```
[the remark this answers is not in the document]
csadorf: Thx for the catch, fixed.

csadorf:
@sphuber @giovannipizzi Are you ok with the state of the AEP or is there anything else needed?

sphuber:
I think there are still a few things to be addressed, or at least I still have some open questions:

1. Can we specify a bit more concretely how we set our dependency requirements in `setup.json`. As I wrote in my original comment something like:  

    For dependencies with at least a v1 we specify only the minor version when using the compatibility operator, unless a specific minor version pin is necessary, due to the incompatibility of the following minor version. For `0.*` dependencies, instead, we specify the patch version which will fix the minor version, since in those stages backwards incompatibility are introduced more readily.

2. As suggested by @ltalirz I think we should add a more concrete plan for generating `requirements.txt`. Do we only include direct or all dependencies?

3. Should we also be a little bit more concrete on how to implement the testing of both abstract and concrete dependencies for a release branch. Not necessarily exact implementation but at least the procedure. Should the whole current test suite be executed for both environments, or just a subset?

4. Should we actually already propose the DM?

ltalirz:
I looked a bit into 2. - in particular using two tools: [pipenv](https://github.com/pypa/pipenv) and [poetry](https://github.com/python-poetry/poetry).
Both tools use their own syntax for specifying dependencies (`Pipfile` for pipenv; `pyproject.toml` for poetry which is at least [PEP 518 conform](https://www.python.org/dev/peps/pep-0518/)) and, for taking full advantage, would need us to switch to their format.
PEP 518 is still provisional and at the moment it is not clear to me which of the competing packaging tools out there will win.

Besides fully switching to one of the new tools, one possibility to address 2. is this:

Use a minimalistic `Pipfile`
```
[dev-packages]

[packages]
aiida-core = {editable = true,path = "."}

[requires]
python_version = "3.7"
````


and generate locked dependencies using
```
pip install pipenv
pipenv lock
pipenv lock --requirements requirements.txt
```
This results in the attached [requirements.txt](https://github.com/aiidateam/AEP/files/4025605/requirements.txt) from which one just needs to remove the  `-e .` dependency.

I think this could be a practical solution for the time being

Note: In the `Pipfile` we specify a python version since we are aiming to provide one environment where aiida-core is guaranteed to work (in pipenv lingo, we are specifying an "application" rather than a "package").

**Comment by ltalirz** [comment]
@csadorf Let me know what you think of the solution proposed above.

I think it's important we move forward with this, since the dependencies in `develop` have already been moved to the more lenient approach suggested here.
Before the next release we should decide how to package a concretized version of the requirements that allow to install a previous version of aiida at some point in the future (e.g. to reproduce a calculation).

**Comment by csadorf** [comment]
I just pushed a few updates to at least partially address your comments.

I've looked more closely into `pipenv` and also the other tools such as `poetry` etc.. Based on my research I must say that I really don't like pipenv's approach and philosophy and the tool is also quite controversial in the wider Python community. I think we should stick to `setup.py` (in combination with `setup.json`) and a `requirements.txt` file and that's it.

For the sake of moving forward, I recommend that we keep the AEP on the slightly higher level as it is right now and refer for details to the wiki. The exact tooling and will need to be adjusted on a more regular basis and I don't want to need to go through an AEP amendment process or whatever every single time for that. Who is filling the role of the DM should also be specified in the wiki.

After we have accepted the AEP, the first task for the DM should be to update all relevant procedures and guidelines related to dependency management within the wiki to conform with the AEP, and then update the CI pipelines according to the wiki. This should be done in collaboration with all AiiDA maintainers and lead developers.

File: 999_dependency_management/readme.md [hunk 2]
```diff
@@ -0,0 +1,147 @@
...
+The DM is responsible for keeping these pages up-to-date and ensuring that tooling and continuous integration pipelines are conform with these procedures.
+
+The CI test pipeline for release branches is to be executed for both specifications.
+The first one to detect unexpected incompatibilities with the latest release version of the dependencies, the second one to test against a well-defined Python environment a user could recreate in an isolated environment.
+
+This approach allows for more flexibility when installing the package into a given environment with either `pip install` or `python setup.py install`, but also provides users an option to install the package with a known set of tested dependencies, *e.g.*, with `pip install -r requirements.txt`.
+
+The dependencies for plugins should be specified in a similar fashion, however dependencies for plugin packages that are likely to be installed in isolation, *i.e.*, have more end-user application character, are encouraged to pin versions per developer discretion.
```
ltalirz: Would be good to give explicit examples here how plugins should depend on aiida-core.
E.g. `aiida-core~=1.0` vs `aiida-core==1.0.1`

ltalirz: Also, what are you actually saying here - are you saying that plugin developers *should* also provide a `requirements.txt`?
I think this suggestion actually applies to the latter case you mention (which, however, is rather unusual).

ltalirz:
```suggestion
AiiDA plugin packages should provide a `setup.py` with dependencies specified as described above, e.g. `aiida-core~=1.0`. 
For packages that have a clear end-user application character, developers may decide to pin the versions of selected dependencies more rigorously -- however, we recommend using a `requirements.txt` for such cases instead.
```

ltalirz:
> I've looked more closely into pipenv and also the other tools such as poetry etc.. Based on my research I must say that I really don't like pipenv's approach and philosophy and the tool is also quite controversial in the wider Python community. I think we should stick to setup.py (in combination with setup.json) and a requirements.txt file and that's it.

I agree - and in the suggestion I made above, we would be sticking to setup.py + requirements.txt - pipenv would just be a tool to get the requirements.txt.
I'm open to alternative routes.

> For the sake of moving forward, I recommend that we keep the AEP on the slightly higher level as it is right now and refer for details to the wiki.

Fine. We still need to implement something in practice, though.

> Who is filling the role of the DM should also be specified in the wiki.

Did you want to volunteer for this originally or do I remember this incorrectly? ;-)
Of course, the manager will have the power to delegate.
As for the AEP, I think it's fine to specify the first DM, but of course this can change down the line.

999_dependency_management/readme.md
csadorf: I think that the same issues that we experience with aiida-core would also be a problem with various plug-ins, so yes, installation dependencies should be only as stringent as needed, the specific environment a plugin was tested in should be specified in the `requirements.txt` file. I can try to clarify that.

csadorf:
> 
> > Who is filling the role of the DM should also be specified in the wiki.
> 
> Did you want to volunteer for this originally or do I remember this incorrectly? ;-)
> Of course, the manager will have the power to delegate.
> As for the AEP, I think it's fine to specify the first DM, but of course this can change down the line.

I'm happy to take this on for now. If I understand correctly, you would like to make that explicit in the AEP?

> > For the sake of moving forward, I recommend that we keep the AEP on the slightly higher level as it is right now and refer for details to the wiki.
> 
> Fine. We still need to implement something in practice, though.

Assuming that I will take on the role as DM for now, then I could start updating the wiki page in parallel while we finalize this AEP.

csadorf:
Unless there are any objections, I will send out a request for comments to the aiida-core developer mailing list by the end of today.

ltalirz:
I would even send it out now, before the AiiDA meeting

csadorf:
@ltalirz This is ready to merge.

ltalirz:
thanks a lot @csadorf !

PR #11, opened 2020-04-17
Efficient object store for the AiiDA repository

giovannipizzi:
 - [x] Status is `submitted`


003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+    a dependency of AiiDA, and is actually the only dependency of the
+    current object-store implementation.
+  - it's quite widespread and so the libary to read and write should be reliable.
+
+- Deletion can just occur efficiently as either a deletion of the loose object, or
+  a removal from the index file (if the object is already packed). Later repacking of the packs can be used to recover
+  the disk space still occupied in the pack files.
+
+- The object store does not need to provide functionality to modify
+  a node. In AiiDA files of nodes are typically only added, and once
```
sphuber, line 229:
```suggestion
  an object. In AiiDA, files of nodes are typically only added, and once
```

espenfl:
Note comments in this issue: https://github.com/aiidateam/aiida-core/issues/335.

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+  `rsync` on a folder with hundreds of thousands of files (not recently accessed) can take minutes or even hours
+  just to check if the files need to be updated, while if the same content is in a single big file, the operation
+  would take much less (possibly even less than a second for a lot of small files). As a note, this is the reason
+  why at the moment we provide a backup script to perform incremental backups of the repository (checking the
+  timestamp in the AiiDA database and only transferring node repository folders of new nodes), but the goal would be to rely instead on standard tools like `rsync`.
+
+
+## Proposed Enhancement 
+The goal of this project is to have a very efficient implementation of an "object store" that:
+- works directly on a disk folder;
```
espenfl, line 54: With "works directly on a disk folder", do you mean that it is compatible with a folder file system as the backbone? And it should as the bare minimum support this?

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+backups).
+
+Here we describe some design criteria for such an object store.
+We also provide an implementation that follows these criteria in the
+[disk-objectstore](https://github.com/giovannipizzi/disk-objectstore)
+repository, where efficiency has also been benchmarked.
+
+### Performance guidelines and requirements
+
+#### Packing and loose objects
```
espenfl, line 80: We should be very careful here. We should try to lean on existing solutions that is used by the object store community.

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+  would take much less (possibly even less than a second for a lot of small files). As a note, this is the reason
+  why at the moment we provide a backup script to perform incremental backups of the repository (checking the
+  timestamp in the AiiDA database and only transferring node repository folders of new nodes), but the goal would be to rely instead on standard tools like `rsync`.
+
+
+## Proposed Enhancement 
+The goal of this project is to have a very efficient implementation of an "object store" that:
+- works directly on a disk folder;
+- ideally, does not require a service to be running in the background,
+  to avoid to have to ask users to run yet another service to use AiiDA;
```
espenfl, line 56: Do we exclude the users that have a proper object store running that is integrated into their workflow? Or do we offer such integrations in a different AEP?

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+given a stream of bytes, stores it as efficiently as possible somewhere, assigns a unique key (a "name") to it,
+and then allows to retrieve it with that key.
+
+The libary also supports efficient bulk write and read operations
+to cover the cases of bulk operations in AiiDA (exporting, importing,
+backups).
+
+Here we describe some design criteria for such an object store.
+We also provide an implementation that follows these criteria in the
+[disk-objectstore](https://github.com/giovannipizzi/disk-objectstore)
```
espenfl, line 75: Can this be integrated with e.g. `rclone`, ref. comments in https://github.com/aiidateam/aiida-core/issues/335?

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+
+- Loose objects are actually written first to a sandbox folder (in the same filesystem),
+  and then moved in place with the correct UUID only when the file is closed, with an atomic operation.
+  This should prevent having leftover objects if the process dies.
+  In this way, we can rely simply on the fact that an object present in the folder of loose objects signifies that the object exists, without needing a centralised place to keep track of them.
+  
+  Having the latter would be a potential performance bottleneck, as if there are many concurrent
+  writers, the object store must guarantee that the central place is kept updated.
+
+- Packing can be triggered by the user periodically, whenever the user wants.
```
espenfl, line 122: What does packing in this context mean? Bundle objects or compression?

003_efficient_object_store_for_repository/readme.md
espenfl: Improving the current performance is nice. But how do we foresee a proper object store integration that gives other benefits? Also, it should be easy to select the repositories we use. E.g. this should in principle be abstracted from AiiDA. We should make sure that if we include a custom light object store like this, that it does not block future direct implementation of a proper object store.

espenfl: One solution would be to follow the principles of e.g. `rclone`. E.g. there is an interface between AiiDA and whatever repositories the interface supports. The `disk-objectstore` should possibly be put into this interface.

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+We have been investigating if existing codes could be used for the current purpose.
+
+However, in our preliminary analysis we couldn't find an existing robust software that was satisfying all criteria.
+In particular:
+
+- existing object storage implementations (e.g. Open Stack Swift, or others that typically provide a
+  S3 or similar API) are not suitable since 1) they require a server running, and we would like to
+  avoid the complexity of asking users to run yet another server, and most importantly 2) they usually work 
+  via a REST API that is extremely inefficient when retrieving a lot of small objects (latencies can
+  be even of tens of ms, that is clearly unacceptable if we want to retrieve millions of objects).
```
espenfl, line 260: This depends wildly on the system setup and how we connect to it. Using cloud solutions is slow, but on-prem object store solutions are indeed rather fast. Of course, they are not designed for performance (but there are solutions, like MinIO) but offer other benefits which would be nice. E.g. access control on object level, replication, integration into larger existing backends at institutions, companies etc. (metadata searches etc.), no file system limit, provides unique id automatically, scalable, high-availability, flexibility and so on. Also, at some point running a large AiiDA infrastructure on regular file systems is going to be a limiting factor, especially in multi user scenarios.

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+  Some examples of issues we envision: managing re-packing (dulwich can only pack loose objects, but not repack existing packs); this can be done via git itself, but then we need to ensuring that when repacking, objects are not garbage-collected and deleted because they are not referenced within git;
+  it's not possible (apparently) to decide if we want to compress an object or not (e.g., in case we want maximum performance at the cost of disk space), but they are always compressed.
+  Also we did not test concurrency access and packing of the git repository, which requires some stress test to assess if it works.
+
+- One possible solution could be a format like HDF5. However we need a very efficient hash-table like access
+  (given a key, get the object), while HDF5 is probably designed to append data to the last dimension of 
+  a multidimensional array. Variables exist, but (untested) probably they cannot scale well
+  at the scale of tens of millions.
+
+- One option we didn't investigate is the mongodb object store. Note that this would require running a server
```
espenfl, line 274: We should leave a comment about using e.g. `rclone` as a common interface.

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+  though (but could be acceptable if for other reasons it becomes advantageous).
+
+Finally, as a note, we stress that an efficient implementation in about 1'000 lines of code has been
```
espenfl, line 277: Fully fledged object store solutions have a very different angle of approach and utility/functionality set and is a different kind of beast. We should be a bit careful how we use the term object store as many users expect much more from that than just being able to bundle files/objects.

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+  and this should be very efficient, even to later retrieve just
+  a subset of the files.
+* Implements compression in a transparent way.
+* It shouldn't be slower than the current AiiDA implementation in writing
+  during normal operation, since objects are stored as loose objects.
+  Actually, it might be faster for nodes without files, as no disk
+  access will be needed anymore.
+
+### Cons
+* Extreme care is needed to convince ourselves that there are no
```
espenfl, line 334: Corner cases typically do not appear until this is battle tested by a larger community in different areas of application. Hence, solutions applied to a larger community are always going to be considered more stable, even though we have 100% test coverage. That only means we test the code that is written, not that the code actually handles all cases.

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+  bugs and no risk of corrupting or losing the users' data.
+* Object metadata must be tracked by the caller in some other database.
+  Similarly, deduplication via SHA hashes is not implemented and need to be
+  implemented by the caller.
+* It is not possible anymore to access directly the folder of a node by opening
```
espenfl, line 339: Yes, maybe this should be a pro instead.

espenfl:
Added a few regular comments to the AEP. This is great and I support this. But it does not really address an integration to a proper object store system. I would still like that to be on the roadmap as a priority if this is implemented.

giovannipizzi:
@greschd I added a note on your comment of automatic repacking in a new commit. You can check 70790e4

I will address the other comments later on

003_efficient_object_store_for_repository/readme.md
```diff
@@ -247,6 +247,26 @@ the different requirements, and represent what can be found in the current imple
   As a note, seeking a file to a given position is what one typically does when watching a 
   video and jumping to a different section.
 
+- Packing in general, at this stage, is left to the user. We can decide (at the object-store level, or probably
```
espenfl, line 250: What does packing here mean? Compression or bundling of several objects/entities? This is a bit difficult to know and I suggest we specify this from get go in the text.

giovannipizzi:
> I think the packing procedures should provide some guarantees (if at all possible) that it will remain in a valid state even if killed. This could of course happen even if the user triggers it manually.
Agreed. This is in general true, but there are very weird corner cases to take into account... You can follow the discussion on these issues [on this issue of the repository](https://github.com/giovannipizzi/disk-objectstore/issues/15).

giovannipizzi:
@sphuber I addressed your comments with a new commit c9f86f7

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+  between the size of each pack (on average, ~4GB/pack for a 1TB repository) and
+  the number of packs (having many packs means that, even when performing bulk access,
+  many different files need to be open, which slows down performance).
+
+- Pack files are just concatenation of bytes of the packed objects. Any new object
+  is appended to the pack (thanks to the efficiency of opening a file for appending).
+  The information for the offset and length of each pack is kept in a single SQLite
+  database for the whole set of objects, as we describe below.
+
+- Packed objects can optionally be compressed. Note that compression is on a per-object level.
```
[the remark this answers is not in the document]
giovannipizzi: No, compression is done fully "transparently" for the user. See new text comments in c9f86f7

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+  It's up to the caller to track this into a filename or a folder structure.
+  The UUID is generated by the implementation and cannot be passed from the outside.
+  This guarantees random distribution of objects in packs, and avoids to have to
+  check for objects already existing (that can be very expensive).
+
+- Pack naming and strategy is not determined by the user. Anyway it would be difficult
+  to provide easy options to the user to customize the behavior, while implementing
+  a packing strategy that is efficient. Moreover, with the current packing strategy,
+  it is immediate to know in which pack to check without having to keep also an index
+  of the packs (this, however, would be possible to implement in case we want to extend the behavior,
```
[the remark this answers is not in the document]
giovannipizzi: Indeed, it should be very easy, and the users wouldn't even notice. Sentence added in c9f86f7

003_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,345 @@
...
+  since anyway we have an index file). But at the moment it does not seem necessary.
+
+- For each object, the SQLite database contains the following fields, that can be considered
+  to be the "metadata" of the packed object: its key (`uuid`), the `offset` (starting
+  position of the bytestream in the pack file), the `length` (number of bytes to read),
+  a boolean `compressed` flag, meaning if the bytestream has been zlib-compressed,
```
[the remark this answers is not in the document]
giovannipizzi: As I now mention in c9f86f7 you can always perform compression yourself (as the user of the library) and store the compressed version. But the advantage of having the OS do it for you is that you don't have to worry about compression. You just store the data you are interested into, and depending on the settings of the repository, or more correctly on the flags with which you will call the packing, packed objects will end up (possibly) being stored as compressed, but you can keep using the object store as usual, and you will always get back the uncompressed version, independent of how the object is actually stored, which compression library was used, ...

On the second point: currently, the compression is zlib, but there are strong suggestions to move to more modern libraries (xz, or snappy, see [e.g. the discussion in this issue](https://github.com/giovannipizzi/disk-objectstore/issues/14))

003_efficient_object_store_for_repository/readme.md
giovannipizzi: Fixed, thanks

003_efficient_object_store_for_repository/readme.md
giovannipizzi: Hi @espenfl , I explained [at the beginning of the section](https://github.com/aiidateam/AEP/pull/11/commits/774c8d426f0a9236e0e78ba1f8f9f4fd81bb690d#diff-34bba23dcb01fde03df9d5e8bfbbcaeeR81-R83):

> The core idea behind the object store implementation is that, instead of writing a lot of small files, these are packed into a few big files, and an index of some sort keeps track of where each object is in the big file. 

Let me know if this is not clear yet.

I will get back to your comments soon.

003_efficient_object_store_for_repository/readme.md
giovannipizzi: I tried to clarify. I mean it directly writes to a folder on a "standard" disk, without the need of going through a running server (as mentioned in the point right below).

003_efficient_object_store_for_repository/readme.md
giovannipizzi: Indeed, maybe the name "object store" is not the correct one - I tried to clarify that what I actually mean is what is referred to as a "key-value" store

003_efficient_object_store_for_repository/readme.md
giovannipizzi: Integration with a proper object store is left out of this AEP, because as I mention (see also discussion below) the actual problem we're trying to solve is latency in retrieving many small files, and actual object stores are much worse than the filesystem on this regard.

Investigation on how to integrate with an "actual" object store can be followed [in this issue](https://github.com/giovannipizzi/disk-objectstore/issues/17).
But unfortunately I don't see how this can be more efficient than just using the disk...

003_efficient_object_store_for_repository/readme.md
giovannipizzi: Clarified

003_efficient_object_store_for_repository/readme.md
giovannipizzi: I think MinIO advertises performance, but this just in the throughput rate (with impressing TB/s). But we're not really interested in that, rather on latency to get a single object. They don't provide any number for that (or at least I couldn't find), but I'd bet is not way better than any other object-store solution, and anyway orders of magnitude slower than getting a file from disk

003_efficient_object_store_for_repository/readme.md
giovannipizzi: Mentioned rclone now

003_efficient_object_store_for_repository/readme.md
giovannipizzi: Mentioned rclone now

003_efficient_object_store_for_repository/readme.md
giovannipizzi: Indeed - I think this is my fault, I should have probably called this a key-value store

003_efficient_object_store_for_repository/readme.md
giovannipizzi: Probably yes!

003_efficient_object_store_for_repository/readme.md
giovannipizzi: I agree - I added a comment on this, and I fully agree that 100% coverage is not enough! I am already aware of a few things that need to be fixed in the code already.

giovannipizzi:
Hi @espenfl I addressed your comments in 5af4c7d

The only (major) think that still needs to be addressed is your comment that this is not an object store. This was due my lack of knowledge to the meaning attached with the work "object store", and I should have called a "key-value store". Before renaming, I'll wait a bit.

Just to clarify: after a lot of consideration, we have realised that:
1. the critical problem now, for AiiDA large profiles, is: storing, backing up, and accessing a lot of small files
2. This AEP aims at solving the issue of having too many small files
3. Unfortunately, object stores are not designed for this goal, and are actually much worse than a filesystem exactly in this respect (accessing a lot of small objects); so they cannot be used as a solution for the specific issue at hand here
4. [I am trying to see in this issue](https://github.com/giovannipizzi/disk-objectstore/issues/17) how to make sure it's easy (at least in read-only mode) to use an object store as a backend to store the files. But due to the intrinsic speed limitations of a real object store, I don't see how one can get good performance there... My most recent findings are that if we only use the repo in read-only mode, we pack everything before putting on the object store, and we are ok with caching on the local machines the actual data, one can get reasonable performance with a relatively easy setup (mount via a S3 interface the bucket with the packed repository, and set a disk caching policy, see details in the linked issue). In this way, we can rely on existing solutions (BOTO, s3fs, clone) to deal with "real" object stores, and we don't have to maintain code for that in AiiDA

Of course, suggestions are welcome, but at the moment the only way to address the critical performance issues that we encounter, with the human resources we have, was to: split the problem, decouple it from the support of "actual" object stores, and try to address the performance issue first (hoping that it can be easily used to support also the object store case (that we didn't forget, but has lower priority with respect to the main issue discussed here), at least for some scenarios like serving a REST API of AiiDA in read-only mode, with data stored once and for all in an object store).

espenfl:
> The only (major) think that still needs to be addressed is your comment that this is not an object store. This was due my lack of knowledge to the meaning attached with the work "object store", and I should have called a "key-value store". Before renaming, I'll wait a bit.

I think we should at least call it something else than object store, or at least specify that in this context we do not mean what the "object store" that the larger community associate with this.

> 
> Just to clarify: after a lot of consideration, we have realised that:
> 
>     1. the critical problem now, for AiiDA large profiles, is: storing, backing up, and accessing a lot of small files

Yes, this I understand and support.

> 
>     3. Unfortunately, object stores are not designed for this goal, and are actually much worse than a filesystem exactly in this respect (accessing a lot of small objects); so they cannot be used as a solution for the specific issue at hand here

In fact, one of the design goals of the object store is precisely to be able to support an infinite number of files. There are also a large number of systems and ways to host and interact with it, so we should be a bit careful putting all solutions into one box. After all high-performance object store is used in high-performance and critical high-availability scenarios that are more complex and intense than what we (maybe ever will) see in AiiDA. But, the problem with these solutions is that they are tailored to a specific task, including the hardware around it. This is a problem; e.g. at this point spend time on solutions in AiiDA that only works (at least to a satisfactory degree) on dedicated hardware and setups. However, in the long term perspective, this is something that anyway needs to be done as the databases and repositories are going to be massive for larger production runs with a lot of partners. But I certainly agree that at this point we should not address this point. However, we should try to avoid introducing concepts now that make such initiatives downstream difficult.

>     4. [I am trying to see in this issue](https://github.com/giovannipizzi/disk-objectstore/issues/17) how to make sure it's easy (at least in read-only mode) to use an object store as a backend to store the files. But due to the intrinsic speed limitations of a real object store, I don't see how one can get good performance there... My most recent findings are that if we only use the repo in read-only mode, we pack everything before putting on the object store, and we are ok with caching on the local machines the actual data, one can get reasonable performance with a relatively easy setup (mount via a S3 interface the bucket with the packed repository, and set a disk caching policy, see details in the linked issue). In this way, we can rely on existing solutions (BOTO, s3fs, clone) to deal with "real" object stores, and we don't have to maintain code for that in AiiDA

In order to gauge performance versus a local or dedicated remote system on fast interconnect one would have to test this when sitting on the storage network. Not sure of the details of the test, but then, the performance is usually rather good and similar to what the parallel file systems can support. And of course, we cannot do better than this, but this goes for all solutions, including the solution that exists today.

Then we have the cases where maximum performance is not necessary an issue, but stability, scalability and longevity is.

> Of course, suggestions are welcome, but at the moment the only way to address the critical performance issues that we encounter, with the human resources we have, was to: split the problem, decouple it from the support of "actual" object stores, and try to address the performance issue first (hoping that it can be easily used to support also the object store case (that we didn't forget, but has lower priority with respect to the main issue discussed here), at least for some scenarios like serving a REST API of AiiDA in read-only mode, with data stored once and for all in an object store).

I support a solution like this if that was unclear from my comments. I am just concerned that this might block future implementations that pursue a true object store implementation (which I suspect have to come at some point). And this in theory also goes past just moving the repo there, but in fact storing the objects themselves in the system, which most likely demands a larger rewrite of some portions of the code. In fact, when using the object store, which anyway relies on a database, it might be that we can remove the AiiDA specific database altogether and rely on the functionality of object store.

In fact, when thinking more about this, this comes down to what kind of users we would like to target; a local setup of AiiDA, or a more integrated approach. The latter would typically be an enterprise solution. Maybe in the future it would make sense to split AiiDA into two parts, one that is more tailored for enterprises (that would need a dedicated setup, support etc.) and one that is more general.

003_efficient_object_store_for_repository/readme.md
espenfl: Ok. Let us think carefully if this can block or make a more proper direct object store integration more complicated in the future. For this it would be nice to have some feedback from someone that works rather deeply with Swift, maybe some people from SwiftStack or similar.

espenfl:
Thanks for addressing each comment @giovannipizzi. I agree.

003_efficient_object_store_for_repository/readme.md
giovannipizzi: I see your point. I think at the moment we want to have at least a good and well performing repository for every user. We can consider as an *optional* extension something requiring a server, for those that require maximal performance. I think that anyway the current API is a step forward, because it introduces a very clear API, that can definitely be applied also to other storage backends (e.g. it provides primitives to store and retrieve data given a key, and also to perform bulk operations).

espenfl: Sounds great.

dev-zero:
I am not entirely sure this belongs here, and even if it does it might come a bit late. If so, please ignore.
What seems to be missing from the comparison of the alternatives is to store the files directly in the PostgreSQL, either as `bytea` or `text`. Even the [PostgreSQL Wiki](https://wiki.postgresql.org/wiki/BinaryFilesInDB#When_is_it_bad_idea_to_store_binary_files_in_the_database.3F) has more or less our use case as an example where it makes sense to store the files in the database (at least the small files). And with [PostgreSQL's TOAST](https://www.postgresql.org/docs/current/storage-toast.html) feature you get the packing and compression for free. On the other end of spectrum (of file sizes) you then have the [Large Objects Feature](https://www.postgresql.org/docs/current/largeobjects.html) which would give streaming access for things like trajectories, wavefunctions, HDF5s, etc.
The big advantage would that backing up (even incrementally) would be covered entirely by [PostgreSQLs WAL archiving and PITR](https://www.postgresql.org/docs/current/continuous-archiving.html) and all the low-level I/O would rely on PostgreSQL, which is fairly robust and safe. The drawback being the complexity when having to mix two possibilities of storing files.

giovannipizzi:
OK - I have updated this AEP (number 006 now, updated the content to the current state).

In particular I  have added a clarification that this is not meant to be a documentation of the `disk-objectstore` package (as the implementation choices might be adapted in the future), but rather as a reference of the reason for the introduction of the package, of the design decisions, and of why they were made.

I have also adapted the text (it referenced and early version of the library, where the key was not the sha256 hash but a random UUID - so I have adapted those parts, or removed the discussion that is now not relevant anymore.

Finally, I have marked the state as "implemented".

I think the best course is to merge this to avoid this information gets lost.

@sphuber or @chrisjsewell could you please review and, if OK, merge? Thanks!

PS @sphuber maybe after this is merged you might want to update and merge also #7? (There is a reference to that PR in this AEP - when you merge you might what to update the link from the PR to the actual AEP number).

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  Also, this should be compared to minutes to hours when storing each object as individual files.
+
+- All operations internally (storing to a loose object, storing to a pack, reading from a loose object or from a pack, compression) should happen via streaming (at least internally, and there should be a public facing API for this).
+  So, even when dealing with huge files, these never fill the RAM (e.g. when reading or writing a multi-GB file, the memory usage has been tested to be capped at \~150MB).
+  Convenience methods are available, anyway, to get directly an object content, if the user wants this for simplicty, and knows that the content fits in RAM.
+
+#### Further design choices
+
+- The key of an object will be its hash.
+  For each container, one can decide which hash algorithm to use; the default one is `sha256` that offers a good compromise between speed and avoiding risk of collisions.
```
sphuber, line 143:
```suggestion
  For each container, one can decide which hash algorithm to use; the default one is `sha256`, which offers a good compromise between speed and avoiding risk of collisions.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  Once an object is stored, it's the responsibility of the `disk-objectstore` library to return the hash of the object that was just stored.
+
+- Using a hash means that we automatically get deduplication of content: if an object is asked to be written, once the stream is received, if the library detects that the object is already present, it will still return the hash key but not store it twice.
+  So, from the point of view of the end application (AiiDA), it does not need to know that deduplication is performed: it just has to send a sequence of bytes, and store the corresponding hash key returned by the library.
+
+- The hashing library can be decided for each container and is configurable; however, for performance reasons, in AiiDA it will be better to decide and stick to one only algorithm.
```
sphuber, line 149:
```suggestion
- The hashing library can be decided for each container and is configurable; however, for performance reasons, in AiiDA it will be better to decide and stick to only one algorithm.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  This will allow to compare e.g. two different repositories (e.g. when sharing data and/or syncing) and establish if data already exists by just comparing the hashes.
+  If different hash algorithms are instead used by the two containers, one needs to do a full data transfer of the whole container, to discover if new data needs to be transfered or not.
+
+- Pack naming and strategy is not determined by the user.
+  Anyway it would be difficult to provide easy options to the user to customize the behavior, while implementing a packing strategy that is efficient.
+  Which object is in which pack is tracked in the SQLite database.
+  But at the moment it does not seem necessary.
```
sphuber, line 156: Is the following really true though? Tracking which object is in which pack _is_ necessary, isn't it?
```suggestion
  Which object is in which pack is tracked in the SQLite database, even though at the moment it does not seem necessary.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  Possible future changes of the internal packing format should not affect the users of the library, since users only ask to get an object by hash key, and in general they do not need to know if they are getting a loose object, a packed object, from which pack, ...
```
sphuber, line 157:
```suggestion
  Possible future changes of the internal packing format should not affect the users of the library, since users only ask to get an object by hash key, and in general they do not need to know if they are getting a loose object, a packed object, from which pack, etc.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+
+- For each object, the SQLite database contains the following fields, that can be considered to be the "metadata" of the packed object: its key (`hashkey`), the `offset` (starting
+  position of the bytestream in the pack file), the `length` (number of bytes to read), a boolean `compressed` flag, meaning if the bytestream has been zlib-compressed, and the `size` of the actual data (equal to `length` if `compressed` is false, otherwise the size of the uncompressed stream, useful for statistics for instance, or to inform the reader beforehand of how much data will be returned, before it starts reading, so the reader can decide to store in RAM the whole object or to process it in a streaming way).
+  In addition, it tracks the number of the pack in which the object is stored.
```
sphuber, line 161:
```suggestion
- For each object, the SQLite database contains the following fields, that can be considered to be the "metadata" of the packed object: its `hashkey`, the `offset` (starting position of the bytestream in the pack file), the `length` (number of bytes to read), a boolean `compressed` flag, meaning if the bytestream has been zlib-compressed, and the `size` of the actual data (which is equal to `length` if `compressed` is false, or the size of the uncompressed stream otherwise).
The latter is useful for statistics for instance, or to inform the reader beforehand of how much data will be returned, before it starts reading, so the reader can decide to store the whole object in RAM or to process it in a streaming way.
  In addition, it tracks the number of the pack in which the object is stored.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+
+- We decided to use SQLite because of the following reasons:
+  - it's a database, so it's efficient in searching for the metadata of a given object;
+  - it does not require a running server;
+  - in [WAL mode](https://www.sqlite.org/wal.html), allows many concurrent readers and one writer, useful to allow to continue normal operations by many Unix processes during repacking;
+  - we access it via the SQLAlchemy library that anyway is already a dependency of AiiDA, and is actually the only dependency of the current object-store implementation;
+  - it's quite widespread and so the libary to read and write should be reliable; in addition, SQLite has a clear [long-term support planning](https://www.sqlite.org/lts.html).
```
sphuber, line 168:
```suggestion
  - it's quite widespread and so the libary to read and write should be reliable;
  - SQLite has a clear [long-term support planning](https://www.sqlite.org/lts.html).
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+
+- Deletion can just occur efficiently as either a deletion of the loose object, or a removal from the index file (if the object is already packed).
```
sphuber, line 170:
```suggestion
- Deletion can be performed efficiently by simply deleting the loose object, or removing the entry from the index file if the object is already packed.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  Later repacking of the packs can be used to recover the disk space still occupied in the pack files.
+  It is hard to find a better strategy that does not require manual repacking but gives all other guarantees especially for fast live operations (as a reference, also essentially all databases do the same and have a "vacuum" operation that is in a sense equivalent to the concept of repacking here).
+
+- The object store does not need to provide functionality to modify an object.
+  In AiiDA, files of nodes are typically only added, and once added they are immutable (and very rarely they can be deleted).
+  
+  If modification is needed, this can be achieved by creation of a new object and deletion of the old one, since this is an extremely rare operation (actually it should never happen).
```
sphuber, line 177:
```suggestion
  If modification is needed, this can be achieved by creation of a new object and deletion of the old one, but this is expected to be a extremely rarely needed operation.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+
+- The current packing format is `rsync`-friendly (that is one of the original requirements).
```
sphuber, line 179:
```suggestion
- The current packing format is `rsync`-friendly, which is one of the original requirements.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  `rsync` has a clever rolling algorithm that can divides each file in blocks and detects if the same block is already in the destination file, even at a different position.
```
sphuber, line 180:
```suggestion
  `rsync` has a clever rolling algorithm that divides each file in blocks and detects if the same block is already in the destination file, even at a different position.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  Therefore, if content is appended to a pack file, or even if a pack is "repacked" (e.g. reordering objects inside it, or removing deleted objects) this does not prevent efficient rsync transfer (this has been tested in the implementation).
+
+- Since it works on a filesystem backend, this would allow to use also other tools, e.g. [`rclone`](https://rclone.org), to move the data to some different backend (a "real" object store, Google Drive, or anything else).
```
sphuber, line 183:
```suggestion
- Since the `disk-objectstore` works on a normal file system, it is possible to use also other tools, e.g. [`rclone`](https://rclone.org), to move the data to some different backend (a "real" object store, Google Drive, or anything else).
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+
+- Appending content to a single file does not prevent the Linux disk cache to work efficiently.
+  Indeed, the caches are per blocks/pages in linux, not per file.
+  Concatenating to files does not impact performance on cache efficiency.
```
sphuber, line 187:
```suggestion
  Concatenating to files does not impact performance of cache efficiency.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  What is costly is opening a file, as the filesystem has to provide some guarantees e.g. on concurrent access.
+  As a note, seeking a file to a given position is what one typically does when watching a video and jumping to a different section.
```
sphuber, line 189: which is efficient? Is that the counterpoint you are trying to make? I would make that explicit then, but not 100% sure that is what you were trying to say
```suggestion
  As a note, seeking a file to a given position is what one typically does when watching a video and jumping to a different section, which is an efficient operation.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+
+- Packing in general, at this stage, is left to the user.
+  We can decide (at the object-store level, or probably better at the AiiDA level) to suggest the user to repack, or to trigger the repacking automatically.
+  This can be a feature introduced at a second time.
+  For instance, the first version we roll out could just suggest to repack periodically in the docs.
+  This could be a good approach, also to bind the repacking with the backups (at the moment, probably backups need to be executed using appropriate scripts to backup the DB index and the repository in the "right order", and possibly using SQLite functions to get a dump).
+
+- As a note: even if repacking is never done by the user, the situation is anyway improved with respect to the current one in AiiDA:
+  - an index of files associated with an AiiDA node will now be stored in the AiiDA DB, so getting the list of files associated to a node without content will not need anymore to access the disk;
```
sphuber, line 198:
```suggestion
  - an index of files associated with an AiiDA node will now be stored in the AiiDA DB, so getting the list of files associated to a node without content will no longer need access to the disk;
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  
+  - there wouldn't be anymore empty folders created for nodes without files;
```
sphuber, line 200:
```suggestion
  - empty folders created for nodes without files will no longer be created;
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  
+  In a second phase, we can print suggestions, e.g. when restarting the daemon, that suggests to repack, for instance if the number of loose objects is too large.
+  We will also provide `verdi` commands to facilitate the user in these maintenance operations.
+
+  Finally, in the future if we are confident that this approach works fine, we can also automate the repacking.
+  We need to be careful that two different processes don't start packing at the same time, and that the user is aware that packing will be triggered, that it might take some time, and that the packing process should not be killed (this might be inconvenient, and this is why I would think twice before implementing an automatic repacking).
+
+### Why a custom implementation of the library
+We have been investigating if existing codes could be used for the current purpose.
+
```
sphuber, line 212:
```suggestion
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+However, in our preliminary analysis we couldn't find an existing robust software that was satisfying all criteria.
+In particular:
+
+- existing object storage implementations (e.g. Open Stack Swift, or others that typically provide a S3 or similar API) are not suitable since 1) they require a server running, and we would like to avoid the complexity of asking users to run yet another server, and most importantly 2) they usually work via a REST API that is extremely inefficient when retrieving a lot of small objects (latencies can be even of tens of ms, that is clearly unacceptable if we want to retrieve millions of objects).
```
sphuber, line 216:
```suggestion
- existing object storage implementations (e.g. Open Stack Swift, or others that typically provide a S3 or similar API) are not suitable since 1) they require a server running, and we would like to avoid the complexity of asking users to run yet another server, and most importantly 2) they usually work via a REST API that is extremely inefficient when retrieving a lot of small objects (latencies can be even of tens of ms, which is clearly unacceptable if we want to retrieve millions of objects).
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+
+- the closest tool we could find is the git object store, for which there exists also a pure-python implementation ([dulwich](https://github.com/dulwich/dulwich)).
```
sphuber, line 218:
```suggestion
- the closest tool we could find is the git object store, for which there exists also a pure Python implementation ([dulwich](https://github.com/dulwich/dulwich)).
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+  We have been benchmarking it, but the feeling is that it is designed to do something else, and adapting to our needs might not be worth it.
+  Some examples of issues we envision: managing re-packing (dulwich can only pack loose objects, but not repack existing packs); this can be done via git itself, but then we need to ensuring that when repacking, objects are not garbage-collected and deleted because they are not referenced within git.
```
sphuber, line 220:
```suggestion
  Some examples of issues we envision: managing re-packing (dulwich can only pack loose objects, but not repack existing packs); this can be done via git itself, but then we need to ensure that when repacking, objects are not garbage-collected and deleted because they are not referenced within git.
```

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,248 @@
...
+- One option we didn't investigate is the mongodb object store.
+  Note that this would require running a server though (but could be acceptable if for other reasons it becomes advantageous).
+
+## Pros and Cons 
+
+### Pros
+* Is very efficient also with hundreds of thousands of objects (bulk reads can read all objects in a couple of seconds).
+* Is rsync-friendly, so one can suggest to use directly rsync for backups instead of custom scripts.
+* A library (`disk-objectstore`) has already been implemented, seems to be very efficient, and is relatively short (and already has tests with 100% coverage, including concurrent writes, reads, and one repacking process, and checking on three platforms: linux, mac os, and windows).
+* It does not add any additional python depenency to AiiDA, and it does not require a service running.
```
sphuber, line 236:
```suggestion
* It does not add any additional Python depenency to AiiDA, and it does not require a service running.
```

006_efficient_object_store_for_repository/readme.md
giovannipizzi: Yes, sorry, a leftover from the first version where object keys were UUIDs and the pack was obtained from the first two characters of the UUID.

006_efficient_object_store_for_repository/readme.md
```diff
@@ -0,0 +1,249 @@
...
+
+- The hashing library can be decided for each container and is configurable; however, for performance reasons, in AiiDA it will be better to decide and stick to only one algorithm.
+  This will allow to compare e.g. two different repositories (e.g. when sharing data and/or syncing) and establish if data already exists by just comparing the hashes.
+  If different hash algorithms are instead used by the two containers, one needs to do a full data transfer of the whole container, to discover if new data needs to be transfered or not.
+
+- Pack naming and strategy is not determined by the user.
+  Anyway it would be difficult to provide easy options to the user to customize the behavior, while implementing a packing strategy that is efficient.
+  Which object is in which pack is tracked in the SQLite database.
+  But at the moment it does not seem necessary.
+  Possible future changes of the internal packing format should not affect the users of the library, since users only ask to get an object by hash key, and in general they do not need to know if they are getting a loose object, a packed object, from which pack, etc.
```
giovannipizzi, line 158:
```suggestion
  Possible future changes of the internal packing format should not affect the users of the library, since users only ask to get an object by hash key, and in general they do not need to know if they are getting a loose object, a packed object, from which pack, etc.
```

giovannipizzi:
Thanks! I've accepted all (just fixed a couple of additional typos I saw). Indeed the sentence saying that it wasn't needed to track which object is in which pack was wrong and I removed it (it was from the old text/very first implementation).

sphuber (approved):

PR #13, opened 2020-07-14
Adopt NEP 29.

csadorf:
 - [x] Status is `submitted`

### DIFF
--- a/003_adopt_nep_29/readme.md
+++ b/003_adopt_nep_29/readme.md
@@ -17,11 +17,12 @@ The [NumPy project](https://numpy.org/) has accepted the [Numpy Enhancement Prop
 
 ## Proposed Enhancement 
 
-This AEP proposed that the AiiDA ecosystem adopts the deprecation policy outlined in NEP 29.
+This AEP proposes that `aiida-core` adopts the Python and NumPy version support schedule and deprecation policy outlined in NEP 29.
+It is strongly recommended that other packages within the AiiDA ecosystem, e.g. plugins, adopt the same policy.
 
 ## Detailed Explanation
 
-After this AEP is accepted, AiiDA will effectively adopt NEP 29 on **2021-01-13** and enter a transition period with a slightly adjusted support schedule until then:
+After this AEP is accepted, `aiida-core` will effectively adopt the policies outlined in NEP 29 on **2021-01-13** and enter a transition period with a slightly adjusted support schedule until then:
 
 | Date           | Python | NumPy | NEP 29 compliant | adjusted
 -----------------|--------|-------|------------------|---------
--- a/003_adopt_nep_29/readme.md
+++ b/003_adopt_nep_29/readme.md
@@ -11,9 +11,9 @@
 
 ## Background 
 
-The aiida-core package is classified as a hybrid between a library and application and dependencies are kept as flexible as possible to allow the installation of the package within various Python software environments in combination with other libraries and applications, see [AEP 2](https://github.com/aiidateam/AEP/blob/master/002_dependency_management/readme.md) for details.
+The `aiida-core` package is classified as a hybrid between a library and application, and dependencies are kept as flexible as possible to allow the installation of the package within various Python software environments in combination with other libraries and applications, see [AEP 2](https://github.com/aiidateam/AEP/blob/master/002_dependency_management/readme.md) for details.
 
-The [NumPy project](https://numpy.org/) has accepted the [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), which recommends that *all projects within the scientific Python ecosystem* adopt a common deprecation policy with respect to the range of minimally supported Python and Numpy versions.
+The [NumPy project](https://numpy.org/) has accepted the [Numpy Enhancement Proposal 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (NEP), which recommends that *all projects within the scientific Python ecosystem* adopt a common deprecation policy with respect to the range of minimally supported Python and Numpy versions.
 
 ## Proposed Enhancement 
 
@@ -21,7 +21,7 @@ This AEP proposed that the AiiDA ecosystem adopts the deprecation policy outline
 
 ## Detailed Explanation
 
-After this AEP is accepted, AiiDA would adopt NEP 29 effectively on **2021-01-13** and enter a transition period with slightly adjusted support schedule until then:
+After this AEP is accepted, AiiDA will effectively adopt NEP 29 on **2021-01-13** and enter a transition period with a slightly adjusted support schedule until then:
 
 | Date           | Python | NumPy | NEP 29 compliant | adjusted
 -----------------|--------|-------|------------------|---------
@@ -32,7 +32,7 @@ After this AEP is accepted, AiiDA would adopt NEP 29 effectively on **2021-01-13
 
 The adjusted support schedule reflects that dropping support for Python 3.5 on an adjusted schedule is not really necessary and schedules the drop of support for Python 3.6 to coincide with the scheduled drop of support for NumPy version 1.16 according to NEP 29 at which point AiiDA's support schedule would be in compliance with NEP 29.
 
-†) The current support of NumPy in aiida-core is defined as `~=1.17<1.18` which means that in order to be fully compliant with NEP 29, AiiDA will need to add support for later versions of NumPy by then.
+†) The current support of NumPy in `aiida-core` is defined as `~=1.17<1.18` which means that in order to be fully compliant with NEP 29, AiiDA will need to add support for later versions of NumPy by then.
 
 ‡) The support for Python 3.6 was [previously announced](http://www.aiida.net/download/) to last until 2021-12-23.
 
@@ -40,10 +40,10 @@ The adjusted support schedule reflects that dropping support for Python 3.5 on a
 
 ### Pros
 
- * The adoption of NEP 29 will harmonize AiiDA's deprecation policy with other major libraries within the Scientific Python ecosystem. As NEP 29 becomes more and more adopted, users will have an expectation of this level of support from AiiDA.
+ * The adoption of NEP 29 will harmonize AiiDA's deprecation policy with other major libraries within the Scientific Python ecosystem. As NEP 29 becomes adopted more broadly, users will have an expectation of the same level of support from AiiDA.
  * The adoption of NEP 29 will reduce the supported range of Python (minor) versions, which will simplify dependency management and testing.
  * The adoption of NEP 29 will remove the need for AiiDA maintainers to maintain their own (custom) support and deprecation schedule.
 
 ### Cons
 
- * The adoption of NEP 29 will slightly reduce the supported range of Python (minor) versions, which might be problematic for users who rely on older Python versions; however it is important to note that other major libraries, such as NumPy, that are likely required in the same context, will also have likely dropped support for these particular versions.
+ * The adoption of NEP 29 will slightly reduce the supported range of Python (minor) versions, which might be problematic for users who rely on older Python versions. However, it is important to note that other major libraries, such as NumPy, that are likely required in the same context, will also have likely dropped support for these particular versions.
--- a/003_adopt_nep_29/readme.md
+++ b/003_adopt_nep_29/readme.md
@@ -13,7 +13,7 @@
 
 The `aiida-core` package is classified as a hybrid between a library and application, and dependencies are kept as flexible as possible to allow the installation of the package within various Python software environments in combination with other libraries and applications, see [AEP 2](https://github.com/aiidateam/AEP/blob/master/002_dependency_management/readme.md) for details.
 
-The [NumPy project](https://numpy.org/) has accepted the [Numpy Enhancement Proposal 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (NEP), which recommends that *all projects within the scientific Python ecosystem* adopt a common deprecation policy with respect to the range of minimally supported Python and Numpy versions.
+The [NumPy project](https://numpy.org/) has accepted the [Numpy Enhancement Proposal 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (NEP 29), which recommends that *all projects within the scientific Python ecosystem* adopt a common deprecation policy with respect to the range of minimally supported Python and NumPy versions.
 
 ## Proposed Enhancement 
 
--- a/003_adopt_nep_29/readme.md
+++ b/003_adopt_nep_29/readme.md
@@ -19,6 +19,23 @@ The [NumPy project](https://numpy.org/) has accepted the [NEP 29](https://numpy.
 
 This AEP proposed that the AiiDA ecosystem adopts the deprecation policy outlined in NEP 29.
 
+## Detailed Explanation
+
+After this AEP is accepted, AiiDA would adopt NEP 29 effectively on **2021-01-13** and enter a transition period with slightly adjusted support schedule until then:
+
+| Date           | Python | NumPy | NEP 29 compliant | adjusted
+-----------------|--------|-------|------------------|---------
+| 2020-09-13     | 3.6+   | 1.17+ | no               | no
+| **2021-01-13** | 3.7+   | 1.17+ | yes†             | **yes**‡
+| 2021-12-26     | 3.8+   | 1.18+ | yes              | no
+
+
+The adjusted support schedule reflects that dropping support for Python 3.5 on an adjusted schedule is not really necessary and schedules the drop of support for Python 3.6 to coincide with the scheduled drop of support for NumPy version 1.16 according to NEP 29 at which point AiiDA's support schedule would be in compliance with NEP 29.
+
+†) The current support of NumPy in aiida-core is defined as `~=1.17<1.18` which means that in order to be fully compliant with NEP 29, AiiDA will need to add support for later versions of NumPy by then.
+
+‡) The support for Python 3.6 was [previously announced](http://www.aiida.net/download/) to last until 2021-12-23.
+
 ## Pros and Cons
 
 ### Pros
--- a/003_adopt_nep_29/readme.md
+++ b/003_adopt_nep_29/readme.md
@@ -1,6 +1,6 @@
 # AiiDA Enhancement Proposal (AEP) Guidelines
 
-| AEP number | XXX                                                          |
+| AEP number | 003                                                          |
 |------------|--------------------------------------------------------------|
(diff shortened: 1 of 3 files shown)


xxx_adopt_nep_29/readme.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Pros
+
+ * The adoption of NEP 29 will harmonize AiiDA's deprecation policy with other major libraries within the Scientific Python ecosystem. As NEP 29 becomes more and more adopted, users will have an expectation of this level of support from AiiDA.
+ * The adoption of NEP 29 will reduce the supported range of Python (minor) versions, which will simplify dependency management and testing.
+ * The adoption of NEP 29 will remove the need for AiiDA maintainers to maintain their own (custom) support and deprecation schedule.
+
+### Cons
+
+ * The adoption of NEP 29 will slightly reduce the supported range of Python (minor) verisons, which might be problematic for users who rely on older Python versions; however it is important to note that other major libraries, such as NumPy, that will likely be required in the same context, will also have dropped support for these particular versions.
```
ltalirz, line 32: Is this even strictly true? Not if one just interprets it as the range of "minimally supported versions" as stated above (although in practice this may be the result ;-) )

csadorf: Happy to rephrase. I think the effect is that other libraries necessary in the same environment are *likely* to have dropped support as well.

csadorf:
@ltalirz I also asked @sphuber , but would you also be willing to champion this AEP?

ltalirz:
Sure. Anyhow, I think this is going to be pretty uncontroversial...

csadorf:
> Sure. Anyhow, I think this is going to be pretty uncontroversial...

Let's hope! :)

xxx_adopt_nep_29/readme.md
csadorf: Rephrased.

sphuber:
> Sure. Anyhow, I think this is going to be pretty uncontroversial...

Well! Here comes some controversy!!

No, just kidding ;) I read the NEP and it seems reasonable to me. Looking at the proposed schedule, they actually deprecate support for minor versions before the actual Python version is EOL. Since we were currently maintaining the latter, we would actually be dropping support earlier, which is less maintenance for us. Like you guys mentioned, the downside of reduced support is justifiable, since the libraries that we depend on, will anyway have dropped support, so our users would likely be forced to upgrade anyway.

Should we just specify as of which time this goes in effect when accepted? We have currently stated that we supported Py 3.5 until EOL which is 2020-09-13. I think we should maybe keep this and from that date start to follow the deprecation rhythm proposed by the NEP. If you agree, could you include this in the AEP? I guess we can say that AiiDA `v1.4` will be the last version to support Py 3.5, which will not be released before 2020-09-13. With the holidays coming up I think two months is not too long to go without minor release.

csadorf:
> Should we just specify as of which time this goes in effect when accepted? We have currently stated that we supported Py 3.5 until EOL which is 2020-09-13. I think we should maybe keep this and from that date start to follow the deprecation rhythm proposed by the NEP. If you agree, could you include this in the AEP? I guess we can say that AiiDA `v1.4` will be the last version to support Py 3.5, which will not be released before 2020-09-13. With the holidays coming up I think two months 
is not too long to go without minor release.

I think the easiest would be to just say that the adoption of the NEP 29 schedule for AiiDA goes into effect after EOL of Python 3.5, i.e., after 2020-09-13. In this way it is unambiguous and we don't restrict our release schedule in any way.

We don't make any particular promises with respect to supported versions of NumPy at the moment, so there we have more flexibility to adopt this schedule even a bit earlier if we want to.

@sphuber Can I add you as champion for this?

sphuber:
> I think the easiest would be to just say that the adoption of the NEP 29 schedule for AiiDA goes into effect after EOL of Python > 3.5, i.e., after 2020-09-13. In this way it is unambiguous and we don't restrict our release schedule in any way.

Fair enough. Anyway, with support, we mean that we will release patch version on the latest minor version supporting that Python version if really necessary. So we could technically already release a hypothetical 1.4 tomorrow that is Python 3.6 and up only.

> @sphuber Can I add you as champion for this?

Sure go ahead

csadorf:
@sphuber I was going to make the above discussed change, however I have trouble finding where exactly we document this:

> We have currently stated that we supported Py 3.5 until EOL which is 2020-09-13.

csadorf:
@ltalirz Could you assign this AEP a number, please?

ltalirz:
Sure, since I think this is likely the next AEP to be merged, let's officially assign number 003.

sphuber:
> @sphuber I was going to make the above discussed change, however I have trouble finding where exactly we document this:

[Python EOL table](https://devguide.python.org/#branchstatus)

If you are talking about where we define _our_ deprecation schedule, I am afraid I cannot find it. I though it used to be on the [roadmap on the wiki](https://github.com/aiidateam/aiida-core/wiki/AiiDA-release-roadmap) but it is no longer there. If it currently isn't documented anywhere, we should define a place. This AEP would be just a good a place as any. We could add a simple link in the wiki roadmap to the AEP once it's merged

ltalirz:
> If you are talking about where we define _our_ deprecation schedule, I am afraid I cannot find it. I though it used to be on the [roadmap on the wiki](https://github.com/aiidateam/aiida-core/wiki/AiiDA-release-roadmap) but it is no longer there. If it currently isn't documented anywhere, we should define a place.

I moved it to http://www.aiida.net/download/ since I wanted new users to see it (who might not necessarily look at the feature roadmap).

csadorf:
> > If you are talking about where we define _our_ deprecation schedule, I am afraid I cannot find it. I though it used to be on the [roadmap on the wiki](https://github.com/aiidateam/aiida-core/wiki/AiiDA-release-roadmap) but it is no longer there. If it currently isn't documented anywhere, we should define a place.
> 

That's what I meant.

> I moved it to http://www.aiida.net/download/ since I wanted new users to see it (who might not necessarily look at the feature roadmap).

So the question is, should we recommend the adjustment of the schedule for Python 3.6 as part of this AEP or do we stick to 23-Dec-2021? I would prefer to adopt NEP 29 after 2020-09-13.

sphuber:
> So the question is, should we recommend the adjustment of the schedule for Python 3.6 as part of this AEP or do we stick to 23-Dec-2021? I would prefer to adopt NEP 29 after 2020-09-13.

I mean, if we accept this AEP, then it would make no sense to not follow the NEP schedule. So I would say yes, we accept this AEP and simply adjust the schedule of dropping Py 3.6 support. This is far enough in the future that this change is justifiable I would say.

csadorf:
@sphuber @ltalirz I've looked at the support schedule in detail and it would make most sense for this to become effective on 2021-01-13, because we are currently only supporting NumPy 1.17+. I've added a detailed explanation for this to the AEP including an adjusted support schedule. Please have a a look.

003_adopt_nep_29/readme.md
```diff
@@ -0,0 +1,49 @@
...
+| Title      | Adopt NEP 29                                                 |
+| Authors    | [Carl Simon Adorf](mailto:simon.adorf@epfl.ch) (csadorf)     |
+| Champions  | [Leopold Talirz](mailto:leopold.talirz@epfl.ch) (ltalirz), [Sebastiaan Huber](mailto:sebastiaan.huber@epfl.ch) (sphuber) |
+| Type       | P - Process                                                  |
+| Created    | 14-Jul-2020                                                  |
+| Status     | submitted                                                    |
+
+## Background 
+
+The aiida-core package is classified as a hybrid between a library and application and dependencies are kept as flexible as possible to allow the installation of the package within various Python software environments in combination with other libraries and applications, see [AEP 2](https://github.com/aiidateam/AEP/blob/master/002_dependency_management/readme.md) for details.
```
sphuber, line 14:
```suggestion
The `aiida-core` package is classified as a hybrid between a library and application, and dependencies are kept as flexible as possible to allow the installation of the package within various Python software environments in combination with other libraries and applications, see [AEP 2](https://github.com/aiidateam/AEP/blob/master/002_dependency_management/readme.md) for details.
```

003_adopt_nep_29/readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+The [NumPy project](https://numpy.org/) has accepted the [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), which recommends that *all projects within the scientific Python ecosystem* adopt a common deprecation policy with respect to the range of minimally supported Python and Numpy versions.
```
sphuber, line 16:
```suggestion
The [NumPy project](https://numpy.org/) has accepted the [Numpy Enhancement Proposal 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (NEP), which recommends that *all projects within the scientific Python ecosystem* adopt a common deprecation policy with respect to the range of minimally supported Python and Numpy versions.
```

003_adopt_nep_29/readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Proposed Enhancement 
+
+This AEP proposed that the AiiDA ecosystem adopts the deprecation policy outlined in NEP 29.
```
sphuber, line 20:
```suggestion
This AEP proposes that `aiida-core` adopts the deprecation policy outlined in NEP 29.
```
Not sure if we can speak for the plugins

003_adopt_nep_29/readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Detailed Explanation
+
+After this AEP is accepted, AiiDA would adopt NEP 29 effectively on **2021-01-13** and enter a transition period with slightly adjusted support schedule until then:
```
sphuber, line 24:
```suggestion
After this AEP is accepted, AiiDA will effectively adopt NEP 29 on **2021-01-13** and enter a transition period with a slightly adjusted support schedule until then:
```

003_adopt_nep_29/readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+†) The current support of NumPy in aiida-core is defined as `~=1.17<1.18` which means that in order to be fully compliant with NEP 29, AiiDA will need to add support for later versions of NumPy by then.
+
+‡) The support for Python 3.6 was [previously announced](http://www.aiida.net/download/) to last until 2021-12-23.
+
+## Pros and Cons
+
+### Pros
+
+ * The adoption of NEP 29 will harmonize AiiDA's deprecation policy with other major libraries within the Scientific Python ecosystem. As NEP 29 becomes more and more adopted, users will have an expectation of this level of support from AiiDA.
```
sphuber, line 43:
```suggestion
 * The adoption of NEP 29 will harmonize AiiDA's deprecation policy with other major libraries within the Scientific Python ecosystem. As NEP 29 becomes adopted more broadly, users will have an expectation of the same level of support from AiiDA.
```

003_adopt_nep_29/readme.md
```diff
@@ -0,0 +1,49 @@
...
+| Date           | Python | NumPy | NEP 29 compliant | adjusted
+-----------------|--------|-------|------------------|---------
+| 2020-09-13     | 3.6+   | 1.17+ | no               | no
+| **2021-01-13** | 3.7+   | 1.17+ | yes†             | **yes**‡
+| 2021-12-26     | 3.8+   | 1.18+ | yes              | no
+
+
+The adjusted support schedule reflects that dropping support for Python 3.5 on an adjusted schedule is not really necessary and schedules the drop of support for Python 3.6 to coincide with the scheduled drop of support for NumPy version 1.16 according to NEP 29 at which point AiiDA's support schedule would be in compliance with NEP 29.
+
+†) The current support of NumPy in aiida-core is defined as `~=1.17<1.18` which means that in order to be fully compliant with NEP 29, AiiDA will need to add support for later versions of NumPy by then.
```
sphuber, line 35:
```suggestion
†) The current support of NumPy in `aiida-core` is defined as `~=1.17<1.18` which means that in order to be fully compliant with NEP 29, AiiDA will need to add support for later versions of NumPy by then.
```

003_adopt_nep_29/readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Pros
+
+ * The adoption of NEP 29 will harmonize AiiDA's deprecation policy with other major libraries within the Scientific Python ecosystem. As NEP 29 becomes more and more adopted, users will have an expectation of this level of support from AiiDA.
+ * The adoption of NEP 29 will reduce the supported range of Python (minor) versions, which will simplify dependency management and testing.
+ * The adoption of NEP 29 will remove the need for AiiDA maintainers to maintain their own (custom) support and deprecation schedule.
+
+### Cons
+
+ * The adoption of NEP 29 will slightly reduce the supported range of Python (minor) versions, which might be problematic for users who rely on older Python versions; however it is important to note that other major libraries, such as NumPy, that are likely required in the same context, will also have likely dropped support for these particular versions.
```
sphuber, line 49:
```suggestion
 * The adoption of NEP 29 will slightly reduce the supported range of Python (minor) versions, which might be problematic for users who rely on older Python versions. However, it is important to note that other major libraries, such as NumPy, that are likely required in the same context, will also have likely dropped support for these particular versions.
```

003_adopt_nep_29/readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Detailed Explanation
+
+After this AEP is accepted, AiiDA would adopt NEP 29 effectively on **2021-01-13** and enter a transition period with slightly adjusted support schedule until then:
+
+| Date           | Python | NumPy | NEP 29 compliant | adjusted
+-----------------|--------|-------|------------------|---------
+| 2020-09-13     | 3.6+   | 1.17+ | no               | no
+| **2021-01-13** | 3.7+   | 1.17+ | yes†             | **yes**‡
+| 2021-12-26     | 3.8+   | 1.18+ | yes              | no
```
sphuber, line 30: Is this really what NEP29 proposes? So they drop Python 3.7 support, a year-and-a-half before its EOL which is at [2023-06-27](https://devguide.python.org/#branchstatus). That seems a bit early, no?

003_adopt_nep_29/readme.md
csadorf: The AEPs are specifically targeted at the AiiDA ecosystem, see [here](https://github.com/aiidateam/AEP/blob/master/000_aep_guidelines/readme.md#background). In so far, I think we *can and should* speak for the plugins. However, obviously not everyone is necessarily able to stick to this schedule, so we could add a sentence to recommend to plugin developers to adopt this support schedule. Would that be acceptable?

003_adopt_nep_29/readme.md
csadorf: The NEP 29 recommends to adopt the following guideline for supported Python versions:
> All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.

Python 3.7 was released on 2018-06-27, that is at that point 3.5 years ago. 

Python has [recently adopted](https://www.python.org/dev/peps/pep-0602/) an updated release cadence of 12 months, which is faster than before. This means that supporting all Python versions that have not reached EOL yet will lead to a much larger range of support. Keep in mind that [EOL is primarily focused on bug fixes and security fixes](https://www.python.org/dev/peps/pep-0602/#year-of-full-support-312-more-years-of-security-fixes). The feature development happens on the latest minor release. By ensuring that every Python minor release is supported for at least 42 months (3.5 years), we find a very good compromise between stability, ability to adopt new language features, and a reduced range of Python versions that need to be tested.

003_adopt_nep_29/readme.md
ltalirz: Yes - I think it is fine to make the distinction explicit, i.e. clarifying that aiida-core will adopt the policy and that we *suggest* the policy as a good practice for the entire ecosystem.

sphuber: Agree with both. As long as we specify that this is what `aiida-core` follows but we strongly advise plugins to follow, that would be great.

003_adopt_nep_29/readme.md
sphuber: Ok, fair enough.

sphuber:
> E.g. what if pymatgen decides to drop support for numpy 1.17 before NEP 29 would recommend to do so?

In these cases, we always have the possibility to put an upper limit on requirements for specific Python versions. Not ideal, but if users really want to keep using older Python versions, that may be the tradeoff they have to accept

sphuber:
Final meta question, if you will. What is the procedure for merging these and updating the final "status" of the AEP. It is currently still marked as "submitted" but essentially when the PR is accepted it should be "accepted". Will you just make that final change before merging once the PR is accepted?

003_adopt_nep_29/readme.md
csadorf: Done.

csadorf:
> Final meta question, if you will. What is the procedure for merging these and updating the final "status" of the AEP. It is currently still marked as "submitted" but essentially when the PR is accepted it should be "accepted". Will you just make that final change before merging once the PR is accepted?

Last time I send out a notice to the mailing list, gave the community about a week to provide feedback, and then we changed the status and merged.

However, I think it would actually be better to merge this PR and keep the status as "submitted". Then we can create PRs with change requests and if everything is resolved I create a PR to change the status to accepted. That would be slightly more in line with the PEP process I believe, where many PEPs are "merged" but their status is not yet "accepted".

csadorf:
> There is one question that remains for me:
> Since aiida-core depends on a number of other packages that could be considered "hybrid", we are a bit dependent on what these packages decide to do.
> 
> E.g. what if pymatgen decides to drop support for numpy 1.17 before NEP 29 would recommend to do so?
> It would be very useful to include with this PR a list of relevant AiiDA dependencies (such as pymatgen, ase, perhaps a few others) and check whether they have adopted NEP 29 (and, if not, possibly suggesting it to them).

I agree with @sphuber that we would need to restrict the versions in that case. Maybe we can add a sentence clarifying the procedure in this case, something like:

> If compliance with the policy outlined in NEP 29 cannot be achieved due to limitations imposed by dependencies, appropriate measures must be taken to remedy the situation, including, but not limited to:
> 1. Placing the affected dependency on the problematic dependency watch list.
> 2. If possible, creating an issue in the associated code repository to recommend the adoption of policies outlines in NEP 29.
> 3. Forking, replacing, or dropping the dependency as a measure of last resort.

I don't want to create that list and make it part of this AEP, because I think that such a list should be maintained as part of the dependency management wiki and it should not really affect whether we adopt this AEP or not.

PR #16, opened 2020-07-16
Set status of AEP 002 to implemented.

csadorf:
AEP 002 has been implemented, because the community has appointed a dependency manager and the recommendations outlined in the AEP are implemented as part of the aiida-core dependency specification and CI workflow.

### DIFF
--- a/002_dependency_management/readme.md
+++ b/002_dependency_management/readme.md
@@ -7,7 +7,7 @@
 | Champions  | [Leopold Talirz](mailto:leopold.talirz@epfl.ch) (ltalirz)    |
 | Type       | P - Process                                                  |
 | Created    | 18-Nov-2019                                                  |
-| Status     | active                                                       |
+| Status     | implemented                                                  |
 
 ## Scope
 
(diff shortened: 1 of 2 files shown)


sphuber:
Is there a place where it is documented what the available values are for the status attribute?

csadorf:
> @csadorf Is there a place where it is documented what the available values are for the status attribute?

The possible values for _status_ are discussed here: https://github.com/aiidateam/AEP/tree/master/000_aep_guidelines#detailed-explanation

sphuber:
Is AEP 000 a special case or should that also be changed to "implemented"?

csadorf:
> Is AEP 000 a special case or should that also be changed to "implemented"?

No idea.

kjappelbaum:
I would set it to "implemented" following a bit how meta-PEPs are treated (https://www.python.org/dev/peps/pep-0001/).
"Some Informational and Process PEPs may also have a status of "Active" if they are never meant to be completed. E.g. PEP 1 (this PEP)." PEP1 is also "active"

csadorf:
It seems to me like we will need to have a meta discussion about this. Honestly, for the process AEPs, _active_ appears a bit more natural to me than _implemented_.

PR #17, opened 2020-07-16
Fix title of AEP 003.

csadorf:
I did not notice that I had not updated the title for AEP 003.


csadorf:
Once we've merged this PR, I'll send out the email to the mailing list.

csadorf:
@sphuber Could you approve this?

PR #21, opened 2020-09-18
Add Archive format AEP

chrisjsewell:
 - [x] Status is `submitted`

### DIFF
--- a/005_exportformat/readme.md
+++ b/005_exportformat/readme.md
@@ -7,7 +7,7 @@
 | Champions  | [Giovanni Pizzi](mailto:giovanni.pizzi@epfl.ch) (giovannipizzi) |
 | Type       | S - Standard                                                     |
 | Created    | 18-Sept-2020                                                     |
-| Status     | submitted                                                        |
+| Status     | draft                                                        |
 
 ## Background
 
--- a/005_exportformat/readme.md
+++ b/005_exportformat/readme.md
@@ -30,7 +30,13 @@ such that:
 * `data.json` contains all requisite information from the SQL Database.
 * `nodes` contains the "object store" files per node, organised by UUID: `xxyy-zz...`.
 
-Particularly for large export archives, writing to (export) and reading from (import) `data.json` represents a significant bottle-neck in performance for these processes, both in respect to memory usage and process speed.
+This format has two principle shortcomings:
+
+1. The file repository is inefficiently stored. 
+   Each file is written as a singular, uncompressed file, which requires a large number of [inode](https://en.wikipedia.org/wiki/Inode) metadata objects and leads large bottlenecks content indexing.
+2. For the validity of `data.json` to be determined, the entire content has to be read into memory, which becomes a limiting factor for export size.
+
+Particularly for large export archives (many millions of nodes), writing to (export) and reading from (import) `data.json` represents a significant bottle-neck in performance for these processes, both in respect to memory usage and process speed.
 
 ## Proposed Enhancement
 
--- a/005_exportformat/readme.md
+++ b/005_exportformat/readme.md
@@ -1,0 +1,103 @@
+# New Export Format
+
+## Hard requirements
+
+* Memory usage As an AiiDA user I want to be able to import/export databases regardless of whether the size of the database is larger than my memory.
+   * E.g. a repository without files and 16GB of attributes/extras: export & import should work with memory of, say, 1-2 GB
+   * How to accomplish this?  Import all nodes, then import all links (all in a transaction), then do checks?
+   * Seb: Currently these checks are done one the frontend level; perhaps it would be faster to do this at the database level
+* Performance As an AiiDA user, I want to be able to import/export 6M nodes in less than 6h, at least in ‘exclusive’ mode.[a]
+   * Note on the exclusive mode: it is going to be possible to achieve this target if we import directly to the pack files. However, it might not be too safe to do it while AiiDA is being used. But I think this is OK: if you really have a huge import to do, most probably it’s the first thing you do after creating a profile (so before starting the daemon or a verdi shell), or it’s OK to stop working for some hours if you get orders of magnitude speedup.
+   * There should still be a ‘default’ mode that imports as loose objects and is safe to use for normal use. HOWEVER, this should WARN the user if there are too many objects (threshold to be decided/tested, e.g. 10’000 or 100’000 objects?), saying that it’s better to stop the daemon, import quickly, and then restart using the daemon.[b]
+* Archive size As an AiiDA user downloading an Archive from Materials Cloud, I expect the size of the archive (after decompression) is of the same order of the size when imported in AiiDA (DB+repo), maximum a factor 1.5x 
+* Disk usage Exporting/importing an archive should not at any point require more free disk space than 2x the size of the uncompressed AiiDA data
+   * i.e., the import shouldn’t create too many temporary copies in the import process
+* Data integrity As an AiiDA user I don't want imports to lead to inconsistencies in my AiiDA database.
+* Data longevity [proposed by Leopold] As an AiiDA user, I want to be able to inspect and reuse the data I export today in 10 (or more) years time  (perhaps even if AiiDA is no longer maintained)
+   * [Leopold] To me, this puts a big emphasis on[c][d][e][f]
+(a) limiting the number of libraries needed to read the export format
+(b) relying only on the most widely used and well established libraries that are likely to be supported for a very long time
+(c) A detailed specification of the format
+   * [Leopold] The old export format does a good job at this - all you need is zip and a JSON parser (you can even open the text files and edit them manually).
+Perhaps a mild modification to the more easily streamable json lines format + a direct connection between the zipped files & the disk object store could be a solution
+   * [Giovanni] See my extensive comment. SQLite is very good for long-term preservation. We should enforce only few (1?) compression algorithm (zlib?)
+   * Any mandatory (python) requirement should be well maintained and easy to install with pip
+Nice to have:
+   * Transactions As an AiiDA user I don't want partial imports in case an import fails
+   * Note: I [GP] think it’s ok if some space gets occupied on disk, e.g. in the import process, but then remains unaccounted for if the import fails (but nodes, instead, SHOULD NOT appear). In this case, there should be simple operations (full repack, or in most cases just a `clean_storage` of the disk-objectstore container) to reclaim that space, similarly to when objects are deleted).
+   * Inspection As a user I want to be able to inspect the content of an export file to know:
+   * how many nodes, computers, users, groups there are (GP: and also how many objects, important to warn the user about the time it will take)
+   * check whether a given set of uuids is in the export file
+   * list all uuids
+   * for developers: given an object hashkeys, know whether it's there
+   * Be able to extract subsets/single nodes from a big export file (or at least single objects)
+   * NOTE: probably this feature will be useful for the progress bar feature
+   * User interface (suggested by Giovanni) As a user, I want to be able to get an interactive progress information (progress bar) that with some approximate accuracy tells me how long the whole process will take (both for imports and for exports)
+   * It’s OK if it’s off by a factor of 2-3[g]
+   * It should be optional[h][i][j]
+   * If activated, shouldn’t make the import or export significantly slower
+   * NOTE: probably this feature will require the possibility to inspect the content of an export file
+   * Delta increments and future push/pull interface (Giovanni) As a user, I would like to be able to easily create export files on one profile to mirror one profile onto another, just transferring the missing nodes/objects.
+   * This will allow for efficient push/pull in the future.
+   * In the end it does not need to be implemented via export files, but if it is it’s easier to write the push/pull code around it, without the need for custom code.
+ 
+## Suggestions
+
+Simon: Consider zarr format - update on hdf5 with better concurrent access
+Gio: Let's take the simplest & most robust solution that fits our needs. zarr seems a bit oriented towards storing n-dimensional arrays.
+Leopold: If the export file contains an sqlite database, then we need to specify for each archive version, which sqlite version (or the disk object store version <> AiiDA version) it is compatible with.[k][l]
+Gio: I suggest to drop the tar.gz format since you have to unzip the entire file to know what it contains
+Gio: We could store node-level jsons in disk object store + a list of uuid<>hashkey
+Leo: If you're already making your export file a database (object store sqlite), why spend so much thought on how to serialize the data from postgres rather than just dumping it into the sqlite database as well?[m][n]
+Leo: Here another interesting resource on possible data formats (e.g. avro) https://www.oreilly.com/library/view/google-bigquery-the/9781492044451/ch04.html#loading_data_efficiently
+   * pro avro: [o]
+pre-defined (JSON) schemas => no per-record overhead; 
+binary => smaller file size, faster reading
+RPC support => can be used a bit like a database
+   * pro json lines: human-readable; simple
+Casper: A regular zip file has intelligent indexing, if I remember correctly, making it easy to utilize for inspection purposes, however, it does not account for the content to zip. I.e., we would probably still have to decide on the “information”/”inner” format.
+
+## Concrete proposals
+
+* JSON lines + files version (easiest to browse without special tools; not the most performant)
+  * store nodes, links, groups in nodes.jsonl, links.jsonl, groups.jsonl JSON lines format
+  * store file repository in a folder structure similar to how it is done now (can be added directly to the zip file without ever creating folder structure on the file system)
+
+* Sqlite/packfiles version
+
+* Hybrid version: create JSON files, but store them in the object store. Just keep a minimal JSON index (e.g. a mapping uuid -> hashkey) - could also be stored in multiline json, even if in many cases one might have to load it all in memory.
+
+## Comments
+
+[a] I know we can never have these requirements very precise and they will always be a vague estimate, but won't this also rely a lot on the machine being used. Is this target of 1Mnodes/hour for a standard laptop, a workstation or a dedicated AiiDA server?
+
+[b] For this requirement, it then becomes crucial that we can really introspect an archive very fast to get a rough sense of the size of the contents. Ideally by scanning on the file, but if this is too slow, we might need to create some "index" file with some stats that our commands typically use. This also goes back to the `verdi export inspect` command and making that as efficient as possible.
+
+[c] Agreed on all three points. However, your confusion (in a private email) that we shouldn't use SQLite does not hold. Indeed, looking at your points: (a) SQLite is in the standard python library, no dependencies required (I use SQLAlchemy because I'm lazy, but I could drop it in disk-objectstore and have 0 dependencies).
+(b) SQLite is a very stable and robust format. They have a dedicated page: https://www.sqlite.org/lts.html, intend to support the format at least until 2050, and SQLite is  the recommended by the US Library Of Congress for the preservation of digital content.
+(c) Indeed, this is important to write down clearly.
+
+The requirement boils down to: given a SHA of an object, get the content. If uncompressed, you just do a SELECT in SQLite, get the offset and the length, and then f=open(pack_file, 'rb'); f.seek(offset); content=f.read(length). If it's compressed, you just need to uncompress with standard libraries (zlib). I agree, though, that we shouldn't use exoteric compression libraries in the export format (OK instead for internal use if wanted).
+
+[d] Great - thanks for checking, I think this makes sqlite a valid choice. 
+However, I do think we still need to weigh it against JSON. Being able to simply open a JSON file in a text editor and manipulate it (even without needing to know anything about the format specification) is an added value that a binary format cannot provide.
+
+And to cite the Library of congress notice on sqlite: "As of this writing (2018-05-29) the only other recommended storage formats for datasets are XML, JSON, and CSV."
+
+[e] As with all these requirements, ultimately they are all tradeoffs. Going for human-readable and introspectable formats will almost guaranteed lead to less efficient formats, both in storage space as well as import/export speed. Although I see your point that you want your data to be usable even after AiiDA is no longer there (people don't like being locked in) I think that realistically this is impractical even with a human-readable format. If you have any reasonably sized archive, although you can directly "read" the data, making any sense out of it without AiiDA will be really difficult. I would therefore be tempted towards a more efficient format even if that means it no longer is directly readable.
+
+[f] Anyway, I think using the disk-objectstore is OK. Look at this issue on how to fetch data without using the library, with (more or less standard) command-line utilities: https://github.com/aiidateam/disk-objectstore/issues/100
+
+[g] Let's not ask the guy who implemented the Windows progress bar to do it then ;)
+[h] Sure, but there should still probably always be a minimum of a progress information given to the user of some kind?
+
+[i] No, you might want to have a quiet mode for some usecases (but indeed, probably verdi export should have some info by default)
+
+[j] Having `-q/--quiet` makes sense, I also simply meant the default. It seems here you want to _not_ make it default, which is fine. I was just clarifying/expanding on this point and its connotations.
+[k] As I mentioned in an earlier comment, the SQLite guys think hard at long-term support - I don't think relying on SQLite will be the problem
+https://www.sqlite.org/lts.html
+[l] As a summary - the SQLite guys are very serious, we can typically trust what they do (their fail-safe mode for DB files written on disk is more robust and has less bugs than postgreSQL!)
+[m] Managing schema migrations become hard, we bind the export file format to the internal AiiDA schema. We should do the opposite (unless it turns out to be impossibly slow) - fix a very clear schema of the export file, aiming at changing it as little as possible, and then map to a schema that we can optimise for speed
+[n] Also, at a node level, it's almost definitely faster to just dump everything in a JSON and write it to disk that to recreate many tables - the DB has to take care of indexes, ACID, avoiding concurrency problems, also when we know nothing of this will happen (e.g. when writing an export file)
+
+[o] If we use Avro, we definitely need to use this library for performance: https://github.com/fastavro/fastavro - still, we should be careful if we need the features of Avro - it's much about schemas, and the largest part of our data will be schema-less - then, just better to dump a JSON per node (and compress it), it will be almost for sure faster.
--- a/005_exportformat/readme.md
+++ b/005_exportformat/readme.md
@@ -95,8 +95,8 @@ Similarly for the object-store, failed imports should not leave large occupied p
 For portability and space management, it is a requirements that the archive be a single (possibly zipped) file.
 This compression should be intrinsic to the format specification, such that read, write and data introspection processes act on and are benchmarked against the zipped archive.
 
-Currently the archive is allowed to be compressed *via* a number of different algorithms (zip, tar.gz).
-For data longevity though, it is desirable that only a single compression algorithm should be enforced.
+Currently the archive is allowed to be compressed *via* two standard compression algorithms (`zip` and `tar.gz`).
+For data longevity, only these standard, well supported compression algorithm should be utilised.
 
 This zip format should also desirably allow for content inspection without fully unzipping the file.
 
--- a/005_exportformat/readme.md
+++ b/005_exportformat/readme.md
@@ -84,7 +84,8 @@ The introspections scan(s) should take an insignificant amount of time, relative
 This introspection could also provide a warning when the default mode is preferable over the exclusive mode, based on a threshold object limit.
 
 Exporting/importing an archive should not at any point require more free disk space than 2x the size of the uncompressed AiiDA data, i.e., the import shouldn’t create too many temporary copies in the import process.
-For example, a repository without files and 16GB of attributes/extras: export & import should not require more than approximately 1-2GB of memory.
+For example, a repository without files and 16GB of attributes/extras: export & import should not require more than approximately 32GB of free disk space.
+Additionaly, the export & import should not require the full archive to be read into memory at any point; both processes should peak at no more than ~2GB of used RAM, irrespective of the database or archive size.
 
 To ensure the data integrity of the final archive or imported database, the interaction with the SQL database should desirably be processed during a single transaction, which can be rolled back in case of import failures.
 Similarly for the object-store, failed imports should not leave large occupied portions of disk space which can not be reclaimed (or at least, if large occupied portions of disk space are left behind, there should be a clear message indicating to the user how to reclaim it, ideally in an efficient way).
@@ -116,14 +117,6 @@ These operation should be very fast and should not be significantly affected by 
 
 It should also be considered how the archive format relates to the internal AiiDA schema, which will likely change over time, with complex schema migrations.
 Ideally the archive format should be independent of this schema, with a well-defined and versioned schema that changes very infrequently.
-
-### Minimal Feature Set
-
-Given the above design requirements, the table below defines a minimal feature set, and compares this to the current implementation.
-
-| Feature                       | Current |
-| ----------------------------- | ------- |
-|                               |         |
 
 ### Proposals
 
--- a/005_exportformat/readme.md
+++ b/005_exportformat/readme.md
@@ -28,7 +28,7 @@ such that:
 
 * `metadata.json` contains the aiida version, export version and other global data.
 * `data.json` contains all requisite information from the SQL Database.
-* `nodes` contains the "objectstore" files per node, organised by UUID: `xxyy-zz...`.
+* `nodes` contains the "object store" files per node, organised by UUID: `xxyy-zz...`.
 
 Particularly for large export archives, writing to (export) and reading from (import) `data.json` represents a significant bottle-neck in performance for these processes.
 
@@ -55,6 +55,10 @@ As an AiiDA user, I want to be able to import/export 6M nodes in less than 6h, a
 
 * Note on the exclusive mode: it is going to be possible to achieve this target if we import directly to the pack files. However, it might not be too safe to do it while AiiDA is being used. But I think this is OK: if you really have a huge import to do, most probably it’s the first thing you do after creating a profile (so before starting the daemon or a verdi shell), or it’s OK to stop working for some hours if you get orders of magnitude speedup.
 * There should still be a ‘default’ mode that imports as loose objects and is safe to use for normal use. HOWEVER, this should WARN the user if there are too many objects (threshold to be decided/tested, e.g. 10’000 or 100’000 objects?), saying that it’s better to stop the daemon, import quickly, and then restart using the daemon.[b]
+
+[a] I know we can never have these requirements very precise and they will always be a vague estimate, but won't this also rely a lot on the machine being used. Is this target of 1M nodes/hour for a standard laptop, a workstation or a dedicated AiiDA server?
+
+[b] For this requirement, it then becomes crucial that we can really introspect an archive very fast to get a rough sense of the size of the contents. Ideally by scanning on the file, but if this is too slow, we might need to create some "index" file with some stats that our commands typically use. This also goes back to the `verdi export inspect` command and making that as efficient as possible.
 
 #### Archive size
 
@@ -105,7 +109,7 @@ Gio: We could store node-level jsons in disk object store + a list of uuid<>hash
 Leo: If you're already making your export file a database (object store sqlite), why spend so much thought on how to serialize the data from postgres rather than just dumping it into the sqlite database as well?[m][n]
 Leo: Here another interesting resource on possible data formats (e.g. avro) https://www.oreilly.com/library/view/google-bigquery-the/9781492044451/ch04.html#loading_data_efficiently
    * pro avro: [o]
-pre-defined (JSON) schemas => no per-record overhead; 
+pre-defined (JSON) schemas => no per-record overhead;
 binary => smaller file size, faster reading
 RPC support => can be used a bit like a database
    * pro json lines: human-readable; simple
@@ -122,10 +126,6 @@ Casper: A regular zip file has intelligent indexing, if I remember correctly, ma
 * Hybrid version: create JSON files, but store them in the object store. Just keep a minimal JSON index (e.g. a mapping uuid -> hashkey) - could also be stored in multiline json, even if in many cases one might have to load it all in memory.
 
 ## Comments
-
-[a] I know we can never have these requirements very precise and they will always be a vague estimate, but won't this also rely a lot on the machine being used. Is this target of 1Mnodes/hour for a standard laptop, a workstation or a dedicated AiiDA server?
-
-[b] For this requirement, it then becomes crucial that we can really introspect an archive very fast to get a rough sense of the size of the contents. Ideally by scanning on the file, but if this is too slow, we might need to create some "index" file with some stats that our commands typically use. This also goes back to the `verdi export inspect` command and making that as efficient as possible.
 
 [c] Agreed on all three points. However, your confusion (in a private email) that we shouldn't use SQLite does not hold. Indeed, looking at your points: (a) SQLite is in the standard python library, no dependencies required (I use SQLAlchemy because I'm lazy, but I could drop it in disk-objectstore and have 0 dependencies).
 (b) SQLite is a very stable and robust format. They have a dedicated page: https://www.sqlite.org/lts.html, intend to support the format at least until 2050, and SQLite is  the recommended by the US Library Of Congress for the preservation of digital content.
--- a/005_exportformat/readme.md
+++ b/005_exportformat/readme.md
@@ -30,10 +30,10 @@ such that:
 * `data.json` contains all requisite information from the SQL Database.
 * `nodes` contains the "object store" files per node, organised by UUID: `xxyy-zz...`.
 
-This format has two principle shortcomings:
+This format has two principal shortcomings:
 
 1. The file repository is inefficiently stored.
-   Each file is written as a singular, uncompressed file, which requires a large number of [inode](https://en.wikipedia.org/wiki/Inode) metadata objects and leads large bottlenecks content indexing.
+   Each file is written as a single, uncompressed file, which requires a large number of [inode](https://en.wikipedia.org/wiki/Inode) metadata objects and leads large bottlenecks content indexing.
 2. For the validity of `data.json` to be determined, the entire content has to be read into memory, which becomes a limiting factor for export size.
 
 Particularly for large export archives (many millions of nodes), writing to (export) and reading from (import) `data.json` represents a significant bottle-neck in performance for these processes, both in respect to memory usage and process speed.
@@ -58,7 +58,7 @@ The following is a list of the key user requirements that a new export format mu
 
 4. Data Integrity: As an AiiDA user I do not want imports to lead to inconsistencies in my AiiDA database.
 
-5. Data Longevitiy: As an AiiDA user, I expect to be able to inspect and reuse the data I export today for at least 10 years into the future (ideally longer).
+5. Data Longevity: As an AiiDA user, I expect to be able to inspect and reuse the data I export today for at least 10 years into the future (ideally longer).
 
 6. Data Introspectability: As an AiiDA user, I expect to be able to obtain rough statistics about the archive, such as the total number of nodes, almost instantaneously.
 
@@ -73,10 +73,10 @@ An initial target though would be approximately one million nodes per hour on a 
 
 The user may run these processes in one of two modes, which may affect the process performance:
 
-* Default mode, can be run whilst the database is being used, i.e. there is one or more daemon running.
+* Default mode, can be run whilst the database is being used, i.e. there is one or more daemon running, and/or one or more verdi shells running or python scripts being executed via `verdi run`.
 * Exclusive mode, requires that the database and file repository are not in use.
 
-In exclusive mode it would be possible to import directly to the new object-store pack files, whereas in default mode the objects would be required to be imported as loose objects.
+In exclusive mode it would be possible to import directly to the new object-store pack files (that has a significant performance boost when importing data involving hundreds of thousands of files in the AiiDA repository or more), whereas in default mode the objects would be required to be imported as loose objects (which is always safe to do even during concurrent access to the repository).
 
 It would also be highly desirable to be able to introspect an archive before import, to feedback an estimate of the process time to the user and provide dynamic progress reporting (e.g. a progress bar).
 This process reporting should aim to be accurate to within a factor of approximately 2-3.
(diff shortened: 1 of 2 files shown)


chrisjsewell:
I'm sure there is room for improvement, but I think this is now in a state for review

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
+# New Export Format
+
+| AEP number | 005                                                              |
+|------------|------------------------------------------------------------------|
+| Title      | Implement a new archive export format                            |
+| Authors    | [Chris Sewell](mailto:christopher.sewell@epfl.ch) (chrisjsewell) |
+| Champions  | [Chris Sewell](mailto:christopher.sewell@epfl.ch) (chrisjsewell) |
```
csadorf, line 7: Please identify a champion for this AEP, it cannot be yourself.

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+* `metadata.json` contains the aiida version, export version and other global data.
+* `data.json` contains all requisite information from the SQL Database.
+* `nodes` contains the "object store" files per node, organised by UUID: `xxyy-zz...`.
+
+Particularly for large export archives, writing to (export) and reading from (import) `data.json` represents a significant bottle-neck in performance for these processes, both in respect to memory usage and process speed.
+
+## Proposed Enhancement
+
+The goal of this project is to first develop a set of agreed requirements for a new archive format, followed by a concrete implementation of the format, and accompanying export and import functions.
```
csadorf, line 37: Please revise the description to focus on the proposed enhancement, we can add detailed notes on its implementation in a later section. This should not be a meta-description of the AEP process itself.

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Detailed Explanation
+
+### User Requirements
+
+The following is a list of the key user requirements that a new export format must address:
+
+1. Process Speed: As an AiiDA user, I want to be able to import/export millions of nodes within hours
```
csadorf, line 45:
```suggestion
1. Process Speed: As an AiiDA user, I want to be able to import/export millions of nodes within hours.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+2. Process Memory Usage: As an AiiDA user, I want to be able to import/export databases regardless of whether the size of the database is larger than my memory.
```
csadorf, line 47:
```suggestion
2. Process Memory Usage: As an AiiDA user, I want to be able to import/export databases regardless of whether the size of the database is larger than the available memory.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+3. Archive Size: As an AiiDA user, downloading an archive from an online repository, I expect the size of the archive (after decompression) is of the same order of the size when imported in AiiDA (DB+repo)
```
csadorf, line 49:
```suggestion
3. Archive Size: As an AiiDA user, downloading an archive from an online repository, I expect the size of the archive (after decompression) to be of the same order as the size when imported in AiiDA (DB+repo).
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+4. Data Integrity: As an AiiDA user I don't want imports to lead to inconsistencies in my AiiDA database.
```
csadorf, line 51:
```suggestion
4. Data Integrity: As an AiiDA user I do not want imports to lead to inconsistencies in my AiiDA database.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+5. Data Accessibility: As an AiiDA user, I want to be able to inspect and reuse the data I export today in 10 (or more) years time
```
csadorf, line 53:
```suggestion
5. Data Accessibility: As an AiiDA user, I expect to be able to inspect and reuse the data I export today for at least 10 years into the future (ideally longer).
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Design requirements
+
+The following design requirements then derive from the above user requirements.
+
+#### Import/Export Process
+
+Naturally, the speed of export/import will be dependant on the the computational hardware under which the process is run.
+An initial target though would be approximately one million nodes per hour, on a standard workstation with four CPU cores.
```
csadorf, line 62:
```suggestion
An initial target though would be approximately one million nodes per hour on a standard workstation with four CPU cores.
```
I think that comma is superfluous.

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+The user may run these processes in one of two modes, which may affect the process performance:
+
+* Default mode, can be run whilst the database is being used, i.e. there is one or more daemons running.
```
csadorf, line 66:
```suggestion
* Default mode, can be run whilst the database is being used, i.e. there is one or more daemon running.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+* Exclusive mode, requires that the database is not in use.
+
+In exclusive mode it would be possible to import directly to the new object-store pack files,
+whereas in default mode the objects would be required to be imported as loose objects.
```
csadorf, line 70:
```suggestion
In exclusive mode it would be possible to import directly to the new object-store pack files, whereas in default mode the objects would be required to be imported as loose objects.
```

csadorf: I think it is necessary to clarify here what the "new object-store pack files" are. Ideally with a reference to the associated AEP.

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+It would also be highly desirable to be able to introspect an archive before import, to feedback an estimate of the process time to the user and provide dynamic progress reporting (e.g. a progress bar). This process reporting should aim to be accurate to within a factor of approximately 2-3.
```
csadorf, line 72:
```suggestion
It would also be highly desirable to be able to introspect an archive before import, to feedback an estimate of the process time to the user and provide dynamic progress reporting (e.g. a progress bar).
This process reporting should aim to be accurate to within a factor of approximately 2-3.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+The introspections scan(s) should take an insignificant amount of time, relative to the actual import/export, and so may necessitate that the archive has a central index, from which to query statistics (see below).
+This introspection could also provide a warning when the default mode is preferable over the exclusive mode, based on a threshold object limit.
+
+Exporting/importing an archive should not at any point require more free disk space than 2x the size of the uncompressed AiiDA data, i.e., the import shouldn’t create too many temporary copies in the import process. For example, a repository without files and 16GB of attributes/extras: export & import should work with memory of approximately 1-2 GB.
```
csadorf, line 76:
```suggestion
Exporting/importing an archive should not at any point require more free disk space than 2x the size of the uncompressed AiiDA data, i.e., the import shouldn’t create too many temporary copies in the import process. 
For example, a repository without files and 16GB of attributes/extras: export & import should not require more than approximately 1-2GB of memory.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+To ensure the data integrity of the final archive or imported database, the interaction with the SQL database should desirably be processed during a single transaction, which can be rolled back in case of import failures.
+Similarly for the object-store, failed imports should not leave large occupied portions of disc space, which can not reclaimed.
```
csadorf, line 79:
```suggestion
Similarly for the object-store, failed imports should not leave large occupied portions of disk space, which can not be reclaimed.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### Archive Accessibility
+
+To mitigate the risk of archives becoming inaccessible, due to future technology deprecations, the following considerations should be made when selecting the archive format,
```
csadorf, line 83:
```suggestion
To mitigate the risk of archives becoming inaccessible, due to future technology deprecations, the following considerations should be made when selecting the archive format:
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+1. The default read/write tool should rely on only a minimal set of well established libraries, with clear long-term support goals.
+2. The format should have a detailed specification, sufficient for third-parties to design independent read/write tools.
+
+The archive format and tools should also provide a standalone means (outside of AiiDA) to inspect the archive file for common information such as:
+
+* The number of database objects, by category (nodes, users, computers, groups, etc)
+* The UUID set contained in the archive
+* The number of object-store objects and potentially a set of object hash keys.
+* Extract subsets/single nodes
```
csadorf, line 93: That item does categorically not belong into this list, but is an additional feature.

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+To mitigate the risk of archives becoming inaccessible, due to future technology deprecations, the following considerations should be made when selecting the archive format,
+
+1. The default read/write tool should rely on only a minimal set of well established libraries, with clear long-term support goals.
+2. The format should have a detailed specification, sufficient for third-parties to design independent read/write tools.
+
+The archive format and tools should also provide a standalone means (outside of AiiDA) to inspect the archive file for common information such as:
+
+* The number of database objects, by category (nodes, users, computers, groups, etc)
+* The UUID set contained in the archive
+* The number of object-store objects and potentially a set of object hash keys.
```
csadorf, line 92:
```suggestion
* the number of database objects, by category (nodes, users, computers, groups, etc)
* the UUID set contained in the archive
* the number of object-store objects and potentially a set of object hash keys
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+* Extract subsets/single nodes
+
+It should also be considered how the archive format relates to the internal AiiDA schema, which will likely change over time, with complex schema migrations.
+Ideally the archive format should be independent of this schema, with a clear schema that changes very infrequently.
```
csadorf, line 96:
```suggestion
Ideally the archive format should be independent of this schema, with a well-defined and versioned schema that changes very infrequently.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+#### Additional Features
+
+An additional feature to consider would be delta increments, such that an existing archive file could be amended during an export.
+This may allow for a push/pull interface for "syncing" an archive file to a particular AiiDA profile.
+
+### Proposals
+
+#### Modification of the current JSON format
+
+The key pro for the current single JSON format is that it is highly stable, future-proof and accessible;
```
csadorf, line 107:
```suggestion
The key advantage of the current single JSON format is that it is highly stable, future-proof and accessible;
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+the file can be opened/read by any standard file editor tool, and directly inspected/edited by the user.
+
+For small archives, this is most probably the best solution.
+However, when considering large archives, single JSON are an extremely poor database format;
```
csadorf, line 111:
```suggestion
However, when considering large archives, single and large JSON-files are an extremely poor database format;
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+For example, node-level JSONs could be stored in the disk object store, together with a minimal index of UUID -> Hashkey mappings.
+At a node level, dumping data into a JSON and writing to disk, would also likely be faster than recreating database tables that must handle indexes, ACID, avoiding concurrency problems, etc.
+When writing an export file, these issues should not necessarily be present.
+
+[JSON streaming](https://en.wikipedia.org/wiki/JSON_streaming) technologies, such as JSONL, also allow for JSON to be streamed, without the need to read the full file into memory.
+It is unclear though if this would actually provide any performance gains, since in many cases the full JSON will still need to be loaded into memory.
+
+#### SQLite database
+
+SQL database storage will likely offer a much more efficient format, both in storage space as well as import/export speed.
```
csadorf, line 125:
```suggestion
SQL database storage will likely offer a much more efficient format, both in storage space requirement as well as import/export speed.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+Of the SQL formats, SQLite likely offers the best solution:
+
+* It is the most simple, being stored as a single file
```
csadorf, line 129:
```suggestion
* It is the most simple, being stored as a single file.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+* It is in the standard Python library, mitigating long-term support issues.
+* It is a very stable and robust format, with a clear long-term support plan until at least until 2050 (see <https://www.sqlite.org/lts.html>)
```
csadorf, line 131:
```suggestion
* It is a very stable and robust format with a clear long-term support plan until at least until 2050 (see <https://www.sqlite.org/lts.html>).
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+The main drawback of using SQL is that it is a binary format and so inherently not directly human-readable.
+The format specification must be known before reading, and also SQLite version should be preserved within the archive.
+
+#### Other database proposals
+
+[Zarr](https://zarr.readthedocs.io) is a relatively new update of the HDF5 format, with better concurrent access.
+However, it appears more oriented towards storing n-dimensional arrays and it long-term stability/support is unclear.
```
csadorf, line 139:
```suggestion
[Zarr](https://zarr.readthedocs.io) is a relatively new standardized format based on the highly stable HDF5 format, with focus on much improved concurrent access.
However, it appears more oriented towards storing n-dimensional arrays and it long-term stability/support is unclear.
```

csadorf, line 139: I do not believe that this single sentence is sufficient grounds for dismissing this technology as a potential solution. Yes, it is relatively young compared to SQLite and HDF5, but there is a standard and the first stable release was in 2016. Yes, both HDF5 and Zarr are designed for storing array-like data, but it can be used as a generic object storage just as well, including, for example, with S3 buckets. Furthermore, it appears to me that the format hits a few other important requirements listed earlier, including the fact that it is indexed, can be used with transparent and selective compression, can be appended to, and is designed for high-performant (parallel) read/write access.

My point is, I am not yet convinced that this technology is the best solution to our problem, but I think we should at least perform a fair and comprehensive evaluation and comparison.

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+[Apache Avro](https://en.wikipedia.org/wiki/Apache_Avro) is another data storage format that utilises JSON, and has some purported benefits:
+
+* Pre-defined (JSON) schemas => no per-record overhead
+* Binary => smaller file size, faster reading
+* Remote Procedure Call (RPC) i support => can be used a bit like a database
```
csadorf, line 145: What does the "i" stand for?

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+Again the long-term stability/support of this format may be an issue.
```
csadorf, line 147:
```suggestion
The long-term stability/support of this format may be an issue.
```

csadorf, line 147: On what grounds?

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### Object-store
+
+The current implementation stores file repositories in a folder structure. These can be added directly to the zip file without ever creating folder structure on the file system.
```
csadorf, line 151:
```suggestion
The current implementation stores file repositories in a folder structure.
These can be added directly to the zip file without ever creating a folder structure on the file system.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+As with the JSON database, this format is highly accessible and stable, but not necessarily performant.
+
+The alternative approach would be to use the newly implemented "packfile" object-store, with coordinating SQLite database.
```
csadorf, line 154:
```suggestion
The alternative approach would be to use the newly implemented "packfile" object-store, in combination with a  SQLite database that contains the index.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+The pros and cons of this approach have been previously assessed in <https://github.com/aiidateam/AEP/pull/11>.
+
+### Archive compression
```
csadorf, line 157: I think that the ability to compress the archive should be discussed as part of the requirements and I would also rephrase it such that it is clear that the "compression" is the requirement, not the fact that one can compress the whole archive. That's always a given.

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+For portability, it is desirable that the full archive be contained within a single zipped file.
```
csadorf, line 159: Again, I think we should rephrase this, because I can *always* place some directories or files in a zip-file.

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
...
+
+Currently the archive is allowed to be compressed *via* a number of different algorithms (zip, tar.gz).
+For data longevity though, it is desirable that only a single compression algorithm should be enforced.
+
+This zip format should also desirably allow for content inspection without fully unzipping the file.
+
+## Pros and Cons
+
+For implementing a new format.
```
csadorf, line 168: It seems to me that this AEP is trying to combine two different arguments: 1. *Whether* a new format should be developed and 2. *What format* that would be. It is easy to combine these two questions by simply making "keep the current format" one of the alternatives.

In that sense, this AEP should identify a proposed specific solution as a result of the earlier discussion and the "Pros and Cons" of that specific solution should be discussed here.

csadorf (commented):
I think this is a good start. I've made a few of suggestions to improve format and content.

005_exportformat/readme.md
```diff
@@ -0,0 +1,180 @@
+# New Export Format
```
chrisjsewell, line 1:
```suggestion
# AiiDA Archive Format
```

005_exportformat/readme.md
chrisjsewell: @giovannipizzi do you want to champion? Is there a specific definition for a champion? Is it just someone else who thinks it is a good idea?

005_exportformat/readme.md
chrisjsewell: just a typo

chrisjsewell, line 145:
```suggestion
* Remote Procedure Call (RPC) support => can be used a bit like a database
```

005_exportformat/readme.md
chrisjsewell: Yeh, I wasn't quite sure what the initial scope and "completeness" of these AEPs had to be to merge in the submitted state?
The question of "whether to keep the current format", as you say, is somewhat tied to already having decided on a specific solution, which in-turn requires that all potential solutions (including Zarr 😉) are fully assessed, and most likely requires some prototyping of one or more of the solutions with tests/benchmarking to compare against the current format etc 

Does all this work need to be completed before a submission is merged? Does it stay as an open PR until this is all done, or is it enough to initially provide an adequate justification for commencing the work?

005_exportformat/readme.md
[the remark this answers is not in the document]
chrisjsewell: changed

005_exportformat/readme.md
chrisjsewell: removed

005_exportformat/readme.md
csadorf: So far we just kept AEPs in the PR stage until they appeared ready. I would be open to merge draft with subsequent updates, but that should probably be discussed in plenum.

005_exportformat/readme.md
```diff
@@ -0,0 +1,182 @@
...
+    ...
+```
+
+such that:
+
+* `metadata.json` contains the aiida version, export version and other global data.
+* `data.json` contains all requisite information from the SQL Database.
+* `nodes` contains the "object store" files per node, organised by UUID: `xxyy-zz...`.
+
+Particularly for large export archives, writing to (export) and reading from (import) `data.json` represents a significant bottle-neck in performance for these processes, both in respect to memory usage and process speed.
```
sphuber, line 33: Maybe we could add here the two principle reasons:

1. The file repository is inefficiently stored. Each file is written as its own file and uncompressed, which requires a lot of inodes and just indexing all file content is slow.
2. For the validity of `data.json` to be determined, the entire content has to be read into memory, which becomes a limiting factor for export size very quickly

005_exportformat/readme.md
```diff
@@ -0,0 +1,182 @@
...
+1. Process Speed: As an AiiDA user, I want to be able to import/export millions of nodes within hours.
+
+2. Process Memory Usage: As an AiiDA user, I want to be able to import/export databases regardless of whether the size of the database is larger than the available memory.
+
+3. Archive Size: As an AiiDA user, downloading an archive from an online repository, I expect the size of the archive (after decompression) to be of the same order as the size when imported in AiiDA (DB+repo).
+
+4. Data Integrity: As an AiiDA user I do not want imports to lead to inconsistencies in my AiiDA database.
+
+5. Data Accessibility: As an AiiDA user, I expect to be able to inspect and reuse the data I export today for at least 10 years into the future (ideally longer).
+
```
sphuber, line 54: I would add:
```
6. Data introspectability: the archive should be able to give rough statistics, such as the total number of nodes or the set of node UUIDs it contains, about the archive's content in an efficient way without requiring the unpacking and/or reading of the entire content

File: 005_exportformat/readme.md [hunk 34]
```diff
@@ -0,0 +1,182 @@
...
+#### Import/Export Process
+
+Naturally, the speed of export/import will be dependant on the the computational hardware under which the process is run.
+An initial target though would be approximately one million nodes per hour on a standard workstation with four CPU cores.
+
+The user may run these processes in one of two modes, which may affect the process performance:
+
+* Default mode, can be run whilst the database is being used, i.e. there is one or more daemon running.
+* Exclusive mode, requires that the database is not in use.
```
sphuber, line 67:
```suggestion
* Exclusive mode, requires that the database and file repository are not in use.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,182 @@
...
+* the number of object-store objects and potentially a set of object hash keys
+* Extract subsets/single nodes
+
+It should also be considered how the archive format relates to the internal AiiDA schema, which will likely change over time, with complex schema migrations.
+Ideally the archive format should be independent of this schema, with a well-defined and versioned schema that changes very infrequently.
+
+#### Additional Features
+
+An additional feature to consider would be delta increments, such that an existing archive file could be amended during an export.
+This may allow for a push/pull interface for "syncing" an archive file to a particular AiiDA profile.
```
sphuber, line 102: I am not sure, but are you referring here to what @giovannipizzi described that he would like to see as a feature? Because I am not sure this is exactly what he meant. What I think he meant was that we should ideally have push/pull functionality on AiiDA databases. And one way of implementing this was through means of export archives. Let's say you have a database, you create an export archive of it and import it in a second db. Then you create additional nodes in db1 and you want to "push" them to db2. The mechanism would be to compute the diff, export those into an archive and then just import that archive in db2. So I don't think we would require directly updating the archives themselves, as this section in the AEP seems to suggest.

005_exportformat/readme.md
```diff
@@ -0,0 +1,182 @@
...
+they must be read in full to access any data and don't support concurrency or ACID (atomicity, consistency, isolation, durability) transactions.
+
+To support this, one could consider extending the current format to move towards a "NoSQL" database type implementation, splitting the JSON into multiple files (see for example [MongoDB](https://en.wikipedia.org/wiki/MongoDB)).
+
+For example, node-level JSONs could be stored in the disk object store, together with a minimal index of UUID -> Hashkey mappings.
+At a node level, dumping data into a JSON and writing to disk, would also likely be faster than recreating database tables that must handle indexes, ACID, avoiding concurrency problems, etc.
+When writing an export file, these issues should not necessarily be present.
+
+[JSON streaming](https://en.wikipedia.org/wiki/JSON_streaming) technologies, such as JSONL, also allow for JSON to be streamed, without the need to read the full file into memory.
+It is unclear though if this would actually provide any performance gains, since in many cases the full JSON will still need to be loaded into memory.
```
sphuber, line 122: What is the advantage of "streamable JSON" if it still requires reading everything in memory :confused:

005_exportformat/readme.md
chrisjsewell: yep, but:

> in an efficient way without requiring the unpacking and/or reading of the entire content

I think this is an implementation rather than user requirements. The user requirement is just that they want to access this data ~instantaneously

005_exportformat/readme.md
chrisjsewell: yep I guess @giovannipizzi can clarify this

005_exportformat/readme.md
chrisjsewell: Perhaps I undersold it here lol. I guess you use it in "searching" for records; e.g. by iterating over each JSON "piece" (loading only that into memory) and checking if it contains the key(s) you are looking for.
I will give a deeper dive on this though

005_exportformat/readme.md
csadorf: Agreed, how it is implemented should not be part of the requirement, however I think what @sphuber is trying to impress is the expected time complexity of that operation. Which should be either O(1) or O(log N). Expressing our "performance" requirements in terms of time complexity first and some order of magnitude time estimate second makes more sense IMO, rather than only providing the latter.

chrisjsewell: Yep some consideration of complexities makes sense 👍

005_exportformat/readme.md
csadorf: I think the outcome of the plenum discussion was to bring this AEP into an agreeable state concerning the requirements as soon as possible and then merge it with the status "submitted" and subsequent updates,, but also @ltalirz is going to have another look at this.

005_exportformat/readme.md
```diff
@@ -0,0 +1,182 @@
+# AiiDA Archive Format
+
+| AEP number | 005                                                              |
+|------------|------------------------------------------------------------------|
+| Title      | Implement a new archive export format                            |
+| Authors    | [Chris Sewell](mailto:christopher.sewell@epfl.ch) (chrisjsewell) |
+| Champions  | [Chris Sewell](mailto:christopher.sewell@epfl.ch) (chrisjsewell) |
```
chrisjsewell, line 7:
```suggestion
| Champions  | [Giovanni Pizzi](mailto:giovanni.pizzi@epfl.ch) (giovannipizzi) |
```

005_exportformat/readme.md
chrisjsewell, line 33:
```suggestion
This format has two principle shortcomings:

1. The file repository is inefficiently stored. 
   Each file is written as a singular, uncompressed file, which requires a large number of [inode](https://en.wikipedia.org/wiki/Inode) metadata objects and leads large bottlenecks content indexing.
2. For the validity of `data.json` to be determined, the entire content has to be read into memory, which becomes a limiting factor for export size.

Particularly for large export archives, writing to (export) and reading from (import) `data.json` represents a significant bottle-neck in performance for these processes, both in respect to memory usage and process speed.
```

005_exportformat/readme.md
chrisjsewell, line 54:
```suggestion
6. Data introspectability: As an AiiDA user, I expect to be able to obtain rough statistics about the archive, such as the total number of nodes, almost instantaneously.

```

005_exportformat/readme.md
```diff
@@ -0,0 +1,182 @@
...
+
+1. Process Speed: As an AiiDA user, I want to be able to import/export millions of nodes within hours.
+
+2. Process Memory Usage: As an AiiDA user, I want to be able to import/export databases regardless of whether the size of the database is larger than the available memory.
+
+3. Archive Size: As an AiiDA user, downloading an archive from an online repository, I expect the size of the archive (after decompression) to be of the same order as the size when imported in AiiDA (DB+repo).
+
+4. Data Integrity: As an AiiDA user I do not want imports to lead to inconsistencies in my AiiDA database.
+
+5. Data Accessibility: As an AiiDA user, I expect to be able to inspect and reuse the data I export today for at least 10 years into the future (ideally longer).
```
chrisjsewell, line 53:
```suggestion
5. Data Longevitiy: As an AiiDA user, I expect to be able to inspect and reuse the data I export today for at least 10 years into the future (ideally longer).
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,182 @@
...
+
+1. The default read/write tool should rely on only a minimal set of well established libraries, with clear long-term support goals.
+2. The format should have a detailed specification, sufficient for third-parties to design independent read/write tools.
+
+The archive format and tools should also provide a standalone means (outside of AiiDA) to inspect the archive file for common information such as:
+
+* the number of database objects, by category (nodes, users, computers, groups, etc)
+* the UUID set contained in the archive
+* the number of object-store objects and potentially a set of object hash keys
+* Extract subsets/single nodes
```
chrisjsewell, line 94:
```suggestion

These operation should be very fast and should not be significantly affected by archive size; degrding in peformance at either O(1) or O(log N) complexity, for N nodes.
```

005_exportformat/readme.md
chrisjsewell: I've added the mention of complexity in the `Archive Accessibility` section below

005_exportformat/readme.md
```diff
@@ -0,0 +1,182 @@
...
+The introspections scan(s) should take an insignificant amount of time, relative to the actual import/export, and so may necessitate that the archive has a central index, from which to query statistics (see below).
+This introspection could also provide a warning when the default mode is preferable over the exclusive mode, based on a threshold object limit.
+
+Exporting/importing an archive should not at any point require more free disk space than 2x the size of the uncompressed AiiDA data, i.e., the import shouldn’t create too many temporary copies in the import process. 
+For example, a repository without files and 16GB of attributes/extras: export & import should not require more than approximately 1-2GB of memory.
+
+To ensure the data integrity of the final archive or imported database, the interaction with the SQL database should desirably be processed during a single transaction, which can be rolled back in case of import failures.
+Similarly for the object-store, failed imports should not leave large occupied portions of disk space, which can not be reclaimed.
+
+#### Archive Accessibility
```
chrisjsewell, line 82:
```suggestion
#### Archive Longevity and Accessibility
```

chrisjsewell:
Points to address from meeting:

- Change Data “accessibility” -> “longevity”
  - Distinguish between longevity and ease of access

- Use cases required for manual changes, but this is normally due to issues in aiida-core caused by “faulty” migrations
  - There should also be some mention/consideration of simplicity vs complexity
  - This also usually relates to low vs high performance, and ascii vs binary

- Move compression to design requirements
  - There should be a mention, in the user requirements, for a single file archive
  - Also relates to space on disc and introspection

- More consideration and explanation of JSON streaming
  - Pro: does not require entire JSON to be kept in memory
  - Likely still needs index mapping

- Derive a minimal required feature set (probably one table per import/export)
  - Compare to current implementation and discuss what features are new/removed
  - One example is storing logs (new)

- Inspecting an archive should not require an AiiDA profile loaded (currently the case)

- API interface should include a `contains_uuid` method

- Benchmarking: 
  Given 1 million nodes, with UUI key + 1 attribute, how long to write and access or check for existence.

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+    ...
+```
+
+such that:
+
+* `metadata.json` contains the aiida version, export version and other global data.
+* `data.json` contains all requisite information from the SQL Database.
+* `nodes` contains the "object store" files per node, organised by UUID: `xxyy-zz...`.
+
+This format has two principle shortcomings:
```
giovannipizzi, line 33:
```suggestion
This format has two principal shortcomings:
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+1. The file repository is inefficiently stored.
+   Each file is written as a singular, uncompressed file, which requires a large number of [inode](https://en.wikipedia.org/wiki/Inode) metadata objects and leads large bottlenecks content indexing.
```
giovannipizzi, line 36:
```suggestion
   Each file is written as a single, uncompressed file, which requires a large number of [inode](https://en.wikipedia.org/wiki/Inode) metadata objects and leads large bottlenecks content indexing.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+1. Process Speed: As an AiiDA user, I want to be able to import/export millions of nodes within hours.
+
+2. Process Memory Usage: As an AiiDA user, I want to be able to import/export databases regardless of whether the size of the database is larger than the available memory.
+
+3. Archive Size: As an AiiDA user, downloading an archive from an online repository, I expect the size of the archive (after decompression) to be of the same order as the size when imported in AiiDA (DB+repo).
+
+4. Data Integrity: As an AiiDA user I do not want imports to lead to inconsistencies in my AiiDA database.
+
+5. Data Longevitiy: As an AiiDA user, I expect to be able to inspect and reuse the data I export today for at least 10 years into the future (ideally longer).
```
giovannipizzi, line 61:
```suggestion
5. Data Longevity: As an AiiDA user, I expect to be able to inspect and reuse the data I export today for at least 10 years into the future (ideally longer).
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+The following design requirements then derive from the above user requirements.
+
+#### Import/Export Process
+
+Naturally, the speed of export/import will be dependant on the the computational hardware under which the process is run.
+An initial target though would be approximately one million nodes per hour on a standard workstation with four CPU cores.
+
+The user may run these processes in one of two modes, which may affect the process performance:
+
+* Default mode, can be run whilst the database is being used, i.e. there is one or more daemon running.
```
giovannipizzi, line 76:
```suggestion
* Default mode, can be run whilst the database is being used, i.e. there is one or more daemon running, and/or one or more verdi shells running or python scripts being executed via `verdi run`.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+* Exclusive mode, requires that the database and file repository are not in use.
+
+In exclusive mode it would be possible to import directly to the new object-store pack files, whereas in default mode the objects would be required to be imported as loose objects.
```
giovannipizzi, line 79:
```suggestion
In exclusive mode it would be possible to import directly to the new object-store pack files (that has a significant performance boost when importing data involving hundreds of thousands of files in the AiiDA repository or more), whereas in default mode the objects would be required to be imported as loose objects (which is always safe to do even during concurrent access to the repository).
```

giovannipizzi: Actually, to be correct, with the current disk-objectstore implementation I think that it would be even OK to add new nodes directly to the pack files while objects are being read (or loose objects being written) by AiiDA processes. What definitely cannot be done is to repack the repo at the same time (but this is an infrequent operation, often controlled by the user). Or, to import directly to packs from two import files at the same time. So - to check, however - we might even evaluate whether to put enough safeguards and activate by default the faster `exclusive` mode.

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+It would also be highly desirable to be able to introspect an archive before import, to feedback an estimate of the process time to the user and provide dynamic progress reporting (e.g. a progress bar).
+This process reporting should aim to be accurate to within a factor of approximately 2-3.
+The introspections scan(s) should take an insignificant amount of time, relative to the actual import/export, and so may necessitate that the archive has a central index, from which to query statistics (see below).
+This introspection could also provide a warning when the default mode is preferable over the exclusive mode, based on a threshold object limit.
+
+Exporting/importing an archive should not at any point require more free disk space than 2x the size of the uncompressed AiiDA data, i.e., the import shouldn’t create too many temporary copies in the import process.
+For example, a repository without files and 16GB of attributes/extras: export & import should not require more than approximately 1-2GB of memory.
```
giovannipizzi, line 87: I think you are mixing disk and memory requirements. Here you probably want to say no more than 32GB of free disk space. Then, as a separate requirement, that we want any import to work on a computer with ~2GB of free memory, to allow e.g. to import also big import files (e.g. 16GB+) in e.g. the Quantum Mobile that has a default of 2GB of RAM allocated to it.

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+To ensure the data integrity of the final archive or imported database, the interaction with the SQL database should desirably be processed during a single transaction, which can be rolled back in case of import failures.
+Similarly for the object-store, failed imports should not leave large occupied portions of disk space, which can not be reclaimed.
```
giovannipizzi, line 90:
```suggestion
Similarly for the object-store, failed imports should not leave large occupied portions of disk space which can not be reclaimed (or at least, if large occupied portions of disk space are left behind, there should be a clear message indicating to the user how to reclaim it, ideally in an efficient way).
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+#### Single File Archive
+
+For portability and space management, it is a requirements that the archive be a single zipped file.
```
giovannipizzi, line 94:
```suggestion
For portability and space management, it is a requirements that the archive be a single (possibly zipped) file.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+This compression should be intrinsic to the format specification, such that read, write and data introspection processes act on and are benchmarked against the zipped archive.
```
giovannipizzi, line 95: I'm not sure it's clear what this sentence is trying to convey.

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+Currently the archive is allowed to be compressed *via* a number of different algorithms (zip, tar.gz).
+For data longevity though, it is desirable that only a single compression algorithm should be enforced.
```
giovannipizzi, line 98: Is this true? I think that if there are more it's OK as long as they are standard and well documented. I agree that however in practice we might want to choose 1, max 2.
The real important requirement is really only the next sentence (This zip format should also desirably allow for content inspection without fully unzipping the file.) to me, the two sentences above are a bit less critical (except maybe saying not to use new/experimental compression algorithms that are not really easy to uncompress with standard tools).

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+1. The default read/write tool should rely on only a minimal set of well established libraries, with clear long-term support goals.
+2. The format should have a detailed specification, sufficient for third-parties to design independent read/write tools.
+
+The archive format and tools should also provide a standalone means (outside of AiiDA) to inspect the archive file for common information such as:
+
+* the number of database objects, by category (nodes, users, computers, groups, etc)
+* the UUID set contained in the archive
+* the number of object-store objects and potentially a set of object hash keys
+
+These operation should be very fast and should not be significantly affected by archive size; degrding in peformance at either O(1) or O(log N) complexity, for N nodes.
```
giovannipizzi, line 115:
```suggestion
These operation should be very fast and should not be significantly affected by archive size, ideally scaling with O(1) or O(log N) complexity, for N nodes.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+### Proposals
+
+#### Modification of the current JSON format
+
+The key advantage of the current single JSON format is that it is highly stable, future-proof and accessible;
+the file can be opened/read by any standard file editor tool, and directly inspected/edited by the user.
+
+For small archives, this is most probably the best solution.
+However, when considering large archives, single and large JSON-files are an extremely poor database format;
+they must be read in full to access any data and don't support concurrency or ACID (atomicity, consistency, isolation, durability) transactions.
```
giovannipizzi, line 137: I would remove the note on ACID - the JSON would be dumped during export by a single process, and never edited, so I don't think this comment applies here.

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+To support this, one could consider extending the current format to move towards a "NoSQL" database type implementation, splitting the JSON into multiple files (see for example [MongoDB](https://en.wikipedia.org/wiki/MongoDB)).
+
+For example, node-level JSONs could be stored in the disk object store, together with a minimal index of UUID -> Hashkey mappings.
```
giovannipizzi, line 141: Small technical note - I think I was suggesting this, but in the end I checked and just writing/reading a multiline JSON is much faster, and if you want to keep an index (that would just increase performance but everything would work even without it) this can just be formed by the byte offset where the line of the JSON of a given node (with given UUID) starts.

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+At a node level, dumping data into a JSON and writing to disk, would also likely be faster than recreating database tables that must handle indexes, ACID, avoiding concurrency problems, etc.
+When writing an export file, these issues should not necessarily be present.
+
+[JSON streaming](https://en.wikipedia.org/wiki/JSON_streaming) technologies, such as JSONL, also allow for JSON to be streamed, without the need to read the full file into memory.
```
giovannipizzi, line 145: As mentioned above, more than streaming, I think a multiline JSON would be a very good technical solution.

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+SQL database storage will likely offer a much more efficient format, both in storage space requirement as well as import/export speed.
+
+Of the SQL formats, SQLite likely offers the best solution:
+
+* It is the most simple, being stored as a single file.
+* It is in the standard Python library, mitigating long-term support issues.
+* It is a very stable and robust format with a clear long-term support plan until at least until 2050 (see <https://www.sqlite.org/lts.html>).
+
+The main drawback of using SQL is that it is a binary format and so inherently not directly human-readable.
+The format specification must be known before reading, and also SQLite version should be preserved within the archive.
```
giovannipizzi, line 160: I would add that 
1. we can limit to SQLite 3 (and readers will know if the format is different)
2. There are GUIs to inspect and edit graphically SQLite files, so this problem is mitigated.

005_exportformat/readme.md
chrisjsewell: As @csadorf rightly pointed out, any folder of files can be turned into a single zipped file.
So the point was to convey that the requirement is not just that the archive can be made a single file, but that it must still be performant to introspect into that single file,
i.e. if you have to perform a timely unzip of the file first, every time you want to query whether it contains a certain UUID, that may not be ideal.

Do you think this is necessary to specify and, if so, how to make it clearer?

chrisjsewell: Haha, I should have read the next comment first, as that was the point I was trying to make.
Maybe this sentence is superfluous then

005_exportformat/readme.md
chrisjsewell: I thought we had agreed only one during the last meeting, but happy to relax this if others are.

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
+# AiiDA Archive Format
+
+| AEP number | 005                                                              |
+|------------|------------------------------------------------------------------|
+| Title      | Implement a new archive export format                            |
+| Authors    | [Chris Sewell](mailto:christopher.sewell@epfl.ch) (chrisjsewell) |
+| Champions  | [Giovanni Pizzi](mailto:giovanni.pizzi@epfl.ch) (giovannipizzi) |
+| Type       | S - Standard                                                     |
+| Created    | 18-Sept-2020                                                     |
+| Status     | submitted                                                        |
```
chrisjsewell, line 10:
```suggestion
| Status     | draft                                                        |
```

README.md
```diff
@@ -14,10 +14,11 @@ ecosystem and to document the decision making process.
 
 | Number | Status           | Title                                                            |
 |--------|------------------|------------------------------------------------------------------|
-| 000    | active           | [AEP guidelines](000_aep_guidelines/readme.md)                   | 
+| 000    | active           | [AEP guidelines](000_aep_guidelines/readme.md)                   |
 | 001    | implemented      | [Drop support for Python 2.7](001_drop_python2/)                 |
 | 002    | implemented      | [AiiDA Dependency Management](002_dependency_management/)        |
 | 003    | active           | [Adopt NEP 29](003_adopt_nep_29/)                                |
+| 005    | submitted        | [New Export Format](005_exportformat/)                           |
```
chrisjsewell, line 21:
```suggestion
| 005    | draft        | [New Export Format](005_exportformat/)                           |
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+It would also be highly desirable to be able to introspect an archive before import, to feedback an estimate of the process time to the user and provide dynamic progress reporting (e.g. a progress bar).
+This process reporting should aim to be accurate to within a factor of approximately 2-3.
+The introspections scan(s) should take an insignificant amount of time, relative to the actual import/export, and so may necessitate that the archive has a central index, from which to query statistics (see below).
+This introspection could also provide a warning when the default mode is preferable over the exclusive mode, based on a threshold object limit.
+
+Exporting/importing an archive should not at any point require more free disk space than 2x the size of the uncompressed AiiDA data, i.e., the import shouldn’t create too many temporary copies in the import process.
+For example, a repository without files and 16GB of attributes/extras: export & import should not require more than approximately 1-2GB of memory.
```
chrisjsewell, line 87:
```suggestion
For example, a repository without files and 16GB of attributes/extras: export & import should not require more than approximately 32GB of free disk space.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,198 @@
...
+Ideally the archive format should be independent of this schema, with a well-defined and versioned schema that changes very infrequently.
+
+### Minimal Feature Set
+
+Given the above design requirements, the table below defines a minimal feature set, and compares this to the current implementation.
+
+| Feature                       | Current |
+| ----------------------------- | ------- |
+|                               |         |
+
```
chrisjsewell, line 127:
```suggestion
```

005_exportformat/readme.md
chrisjsewell: done

005_exportformat/readme.md
```diff
@@ -0,0 +1,191 @@
...
+To ensure the data integrity of the final archive or imported database, the interaction with the SQL database should desirably be processed during a single transaction, which can be rolled back in case of import failures.
+Similarly for the object-store, failed imports should not leave large occupied portions of disk space which can not be reclaimed (or at least, if large occupied portions of disk space are left behind, there should be a clear message indicating to the user how to reclaim it, ideally in an efficient way).
+
+#### Single File Archive
+
+For portability and space management, it is a requirements that the archive be a single (possibly zipped) file.
+This compression should be intrinsic to the format specification, such that read, write and data introspection processes act on and are benchmarked against the zipped archive.
+
+Currently the archive is allowed to be compressed *via* a number of different algorithms (zip, tar.gz).
+For data longevity though, it is desirable that only a single compression algorithm should be enforced.
```
chrisjsewell, line 99:
```suggestion
For data longevity, only stanard, well supported compression algorithm should be utilised.
```

005_exportformat/readme.md
```diff
@@ -0,0 +1,191 @@
...
+
+To ensure the data integrity of the final archive or imported database, the interaction with the SQL database should desirably be processed during a single transaction, which can be rolled back in case of import failures.
+Similarly for the object-store, failed imports should not leave large occupied portions of disk space which can not be reclaimed (or at least, if large occupied portions of disk space are left behind, there should be a clear message indicating to the user how to reclaim it, ideally in an efficient way).
+
+#### Single File Archive
+
+For portability and space management, it is a requirements that the archive be a single (possibly zipped) file.
+This compression should be intrinsic to the format specification, such that read, write and data introspection processes act on and are benchmarked against the zipped archive.
+
+Currently the archive is allowed to be compressed *via* a number of different algorithms (zip, tar.gz).
```
chrisjsewell, line 98:
```suggestion
Currently the archive is allowed to be compressed *via* two standard compression algorithms (`zip` and `tar.gz`).
```

chrisjsewell:
I believe I've now addressed all comments, so merging this in "draft" status (see #23)

PR #23, opened 2020-11-01
Update AEP 000: Add `draft` status

### LINKED ISSUE #22: revision of AEP "status" codes
We've started out with the status codes used also by the jupyter project:

  * `submitted` - this should be the initial status when submitting the pull request to the AEP repository
  * `active` - this AEP has been accepted and people are actively discussing and implementing it
  * `implemented` - this AEP has been implemented
  * `postponed` - this AEP is no longer active, might be interesting for the project but has noone willing to champion it
  * `rejected` - this AEP has been rejected and will not be implemented
  * `withdrawn` - this AEP has been withdrawn by the submitter but can be re-submitted if someone is willing to champion 

In the context of discussions concerning PRs https://github.com/aiidateam/AEP/pull/11 and https://github.com/aiidateam/AEP/pull/21 a question arose about the right time to merge a PR:
Does it make sense to merge a PR once people have agreed to move forward in this direction or should a PR be merged only once all implementation details have been figured out (which, effectively, is *after* its implementation in `aiida-core`)?

The general feeling seemed to be that it would be useful to merge PRs in a draft stage as an intermediate form of saying "we agree that we want this to happen".
As it turns out, this also came up in the jupyter project https://github.com/jupyter/governance/issues/13 (but has so far gone stale).

Alternatives for inspiration:
 * ECMAScript (JavaScript): https://tc39.es/process-document/
 * zeromq: https://github.com/jupyter/enhancement-proposals/issues/11#issuecomment-181046207

ltalirz:
Fixes #22 

Especially with larger projects it can take a significant amount of time
(easily a couple of months) between first discussions on an enhancement
proposal and its actual implementation/release in aiida-core.  With the
current approach, we have typically seen some initial discussion in a
pull request, which then died at some point.

Since there is no "checkpointing", when looking back at the PR after a
long time, it can be unclear what was actually agreed upon and which
questions were left open. Agreeing to merge the PR in a draft state
provides such a checkpoint (besides making the draft more visible
in the AEP repository).

We also remove the 'submitted' status, since it was used only during
the pull request stage and always needed to be changed before merging.


000_aep_guidelines/readme.md
```diff
@@ -40,10 +41,10 @@ A table containing:
     * `S - Standard Track AEP`: describes new features or changes to AiiDA
     * `I - Informational AEP`: describes design issues or best practice. A notable informational PEP is [the Zen of Python](https://www.python.org/dev/peps/pep-0020/)
     * `P - Process AEP`: describes changes to processes in the AiiDA ecoystem, such as a change to the decision making or development process (e.g. new standards for commit messages)
-  * creation date
+  * date of creation
+  * date of last modification
   * status: one of
-    * `submitted` - this should be the initial status when submitting the pull request to the AEP repository
-    * `active` - this AEP has been accepted and people are actively discussing and implementing it
+    * `draft` - this AEP has been accepted, and people are actively discussing and implementing it (further modifications may be necessary)
```
csadorf, line 47:
```suggestion
    * `draft` - this AEP has been accepted for further discussion and development
```
I don't think we should give the impression that a merged AEP in "draft" stage is "ready for implementation" or has received some kind of preliminary "go ahead". I would also be in favor that it still possible to reject a "draft AEP". I was under the assumption from earlier discussion that the main purpose of adding this state was to allow for AEP's to be merged so that they don't linger in PR stage limbo. I would be much less comfortable accepting an AEP into "draft" stage if this means that I can no longer ultimately reject the AEP.

IMO, whether folks work on some kind of prototype implementation in the meantime is slightly out of scope for this process (that's a management decision). For example, many PEPs are accompanied by prototype implementation to work out feasibility and implementation details that are otherwise hard to anticipate, but that certainly does not mean that those implementations can be expected to be merged.

csadorf (changes requested):

000_aep_guidelines/readme.md
ltalirz: Right, I agree I should add add another review stage in between 'draft' and 'implemented'.

I'm happy to reword the description of the draft state as well, but I think it needs to reflect the fact that the gist of the AEP has been endorsed, i.e. that someone working on fleshing out the implementation of this AEP can be confident that it will be merged eventually (perhaps with changes, but not outright rejected).

greschd: >can be confident that it will be merged eventually

I wouldn't put it quite so strongly: Yes, the accepted draft stage means that the AEP has been endorsed, and in the common case should not be rejected.

There can of course be exceptions, where during implementation we discover a significant drawback to the suggested change, that will mean the AEP still has to be rejected.

I would consider merging of a "draft" AEP as a "statement of intent" to accept the implemented AEP. A good signal that it will eventually be merged, but not a guarantee.

Of course, if the implementation of an AEP uncovers a serious flaw in our initial thinking, I also wouldn't consider that effort wasted - after all, this means that it has significantly advanced our understanding of the design space.

ltalirz: Sounds good to me; now we just need to find a compact sentence to express this

greschd: "this AEP has been conditionally approved; it is actively being developed, and its final version is intended to be accepted barring strong reasons for rejecting"?

csadorf: > "this AEP has been conditionally approved; it is actively being developed, and its final version is intended to be accepted barring strong reasons for rejecting"?

That's too strong IMO. Any rejection should be backed by "strong reasons" anyways. Here is my proposal:

> This AEP has been approved for further discussion and the current state reflects the state of decisions of all parties involved.

If the discussion was well-balanced, this intrinsically means that an acceptance of AEP can be expected, but there is no reason to make it that explicit. I think we are trying to coerce governance into the state description, whereas the decision making process of AEPs hasn't even been fully clarified. Who has the authority to accept or reject AEPs? So far we've put AEPs up for discussion with the wider community and then accepted them as long as nobody voiced any concerns.

I also don't think we should be explicit on whether an AEP is implemented or not. Depending on the nature of an AEP it may or may not make sense to "work on it" prior to a final decision.

ltalirz: My motivation for starting this discussion on the description of the draft status was that I would prefer a sentence that gives an interested developer (think e.g. an AiiDA plugin developer; a talented GSOC student; ...) some level of confidence  that if they are volunteering their precious time to work on this, it is not going to be wasted.

If I imagine myself in this position, `this AEP has been approved for further discussion` does not give me this confidence.

In my view, merging an AEP in draft state should be seen as an endorsement of this proposal by the AiiDA team, with the goal of merging an eventual implementation (to be hashed out further with the team) into AiiDA core.


> the decision making process of AEPs hasn't even been fully clarified. Who has the authority to accept or reject AEPs?

The current procedure is described in https://github.com/aiidateam/AEP/tree/master/000_aep_guidelines#how-to-submit-an-aep
It could certainly be narrowed down in more detail (e.g. currently, a PR requires 2 reviews from code owners, i.e. AiiDA team); the necessity to do this has just not arison so far.

ltalirz: Proposed sentence: 
`this AEP is endorsed by the AiiDA team but requires further discussion and implementation`

csadorf: > My motivation for starting this discussion on the description of the draft status was that I would prefer a sentence that gives an interested developer (think e.g. an AiiDA plugin developer; a talented GSOC student; ...) some level of confidence that if they are volunteering their precious time to work on this, it is not going to be wasted.

Quite frankly, I don't think that situation would occur in practice and this preliminary endorsement feels premature. To me AEPs are primarily a tool to document a design and then secondary a decision (approval/rejection) process. The only criterion to merge an AEP in draft stage should be whether it is generally applicable, has the right format, is of sufficient quality, and is considered "worth of a discussion" by the AiiDA team. Anything beyond that would make me personally more hesitant to merge an AEP which means we end up where we started (all drafts remain at the PR stage).

ltalirz:
Given that AEP 005 was already merged in `draft` status and we're currently lacking a formal description of it, let's move forward with this.

I've accepted the suggestion by @csadorf for the description of the `draft` status; I no longer feel as strongly about the wording as I did before, and I think is a good step forward.

ltalirz:
@sphuber please have a look; also pinged @csadorf

000_aep_guidelines/readme.md
```diff
@@ -41,7 +41,7 @@ A table containing:
     * `S - Standard Track AEP`: describes new features or changes to AiiDA
     * `I - Informational AEP`: describes design issues or best practice. A notable informational PEP is [the Zen of Python](https://www.python.org/dev/peps/pep-0020/)
     * `P - Process AEP`: describes changes to processes in the AiiDA ecoystem, such as a change to the decision making or development process (e.g. new standards for commit messages)
-  * date of creation
+  * date of creation: when the pull request for this AEP was opened
```
ltalirz, line 44: One could also use the date when the PR was merged, but I think the more interesting thing to know is how long this idea has been floating around 
+ it's how the date field has been used so far 
+ it requires less work since that way people put it in right from the beginning of the PR

000_aep_guidelines/readme.md
```diff
@@ -40,10 +41,10 @@ A table containing:
     * `S - Standard Track AEP`: describes new features or changes to AiiDA
     * `I - Informational AEP`: describes design issues or best practice. A notable informational PEP is [the Zen of Python](https://www.python.org/dev/peps/pep-0020/)
     * `P - Process AEP`: describes changes to processes in the AiiDA ecoystem, such as a change to the decision making or development process (e.g. new standards for commit messages)
-  * creation date
+  * date of creation: when the pull request for this AEP was opened
+  * date of last modification
```
sphuber, line 45:
```suggestion
  * date of last modification: should be updated whenever an implemented AEP is updated
```

README.md
```diff
@@ -9,12 +9,13 @@ Accepted AEPs can be easily browsed and read on the [online documentation](https
 
 ## Index of AEPs
 
+<<<<<<< HEAD
```
sphuber, line 12:
```suggestion
```

000_aep_guidelines/readme.md
sphuber, line 45: Should we maybe have one more date that mentions the date when it was merged? May be overcomplicating things, but if there is a lot of time between the opening of PR and it being merged as draft for the first time, it might be useful.

It can be left empty when a PR is opened, but then as a last step before merging a new AEP, would be to update the AEP number, and set the merge and last modification date to the day of merging. Could be part of the "merging protocol".

004_calcjob_importer/readme.md
```diff
@@ -180,4 +181,4 @@ However, there are pros and cons in the proposed design of the user interface, w
   I actually think this will not be a problem but I am including it here as it has been brought up as a con of the current design by other in private conversations.
   Since one actually have to construct a `RemoteData` and include it in the inputs, I do not think that this happens by accident or to an unexpecting user.
 * By purposefully not providing a basic interface for the conversion of output files to inputs nodes in `aiida-core`, we run the risk that the various solutions and their interfaces, that will be designed and provided by the plugins will be wildly disparate.
-  This might negate the benefits of the consistent import interface itself in `aiida-core`, by having differing interfaces to create the input nodes themselves, which is a prerequisite step.
\ No newline at end of file
+  This might negate the benefits of the consistent import interface itself in `aiida-core`, by having differing interfaces to create the input nodes themselves, which is a prerequisite step.
```
sphuber, line 184: Why is this line added? It seems irrelevant no?

sphuber (changes requested):
Thanks @ltalirz , phrasing is fine for me. Just some small corrections to make and wondering if we should have a date that records day of merging the AEP, i.e., when it became accepted as a draft

004_calcjob_importer/readme.md
csadorf: Yeah, look like this change made it into wrong commit.

000_aep_guidelines/readme.md
csadorf: I feel like we should try to avoid making things more complicated here. I am already a bit hesitant about the "date of modification" considering that this already well-documented as part of the revision history and likely something that we would forget to update very regularly.

The whole AEP process is already quite bureaucratic, I would recommend to keep "procedures" and "protocols" as simple as possible.

csadorf (approved):
I'm going to approve this on the condition that the commit is cleaned up (@sphuber pointed out some artifacts). Other than that, I think this is fine as-is. I would recommend against adding the "date of modification" field, but I don't feel strongly about it and would find it more important that this PR is finally merged.

000_aep_guidelines/readme.md
sphuber: I was also not to sure about this, just wondering. So fine to leave this out.

ltalirz: I'll remove it

004_calcjob_importer/readme.md
ltalirz: The line isn't added, it's already there - I don't see from github what the difference is, perhaps related to spaces by some editor. I guess it's fine.

ltalirz:
@sphuber Would you like to do a final check?

004_calcjob_importer/readme.md
sphuber: weird, ok then

sphuber (approved):

PR #29, opened 2021-12-07
AEP 009: Move materials science data types out of AiiDA core

mbercx:
AEP to discuss moving the materials science related data types out of `aiida-core`, taking this opportunity to redesign them.

 - [X] Used [AEP template](../blob/master/000_aep_guidelines/readme.md) from AEP 0
 - [X] Status is `submitted`
 - [X] Added type & status labels to PR

### DIFF
--- a/009_move_matsci_data/readme.md
+++ b/009_move_matsci_data/readme.md
@@ -1,0 +1,108 @@
+# Move materials science data types out of AiiDA core
+
+| AEP number | 009                                                          |
+|------------|--------------------------------------------------------------|
+| Title      | Move materials science data types out of AiiDA core          |
+| Authors    | [Marnik Bercx](mailto:marnik.bercx@epfl.ch) (mbercx)         |
+| Champions  | [Marnik Bercx](mailto:marnik.bercx@epfl.ch) (mbercx)         |
+| Type       | S - Standard Track AEP                                       |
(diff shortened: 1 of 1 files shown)


mbercx:
Currently the (rough) draft AEP aims to discuss both _how_ to move the data types out of AiiDA core, as well as what changes to make to the data types. This may be a bit ambitious, however, and it would be easier to collaborate on the design of the new data types in the new repository. So perhaps we should limit the scope of this AEP to discussing:

* Where to move the data types (one new repository or multiple).
* How to execute the move, i.e. deprecation and migration.

If we agree on this, I'll move out the issues that I've triaged so far for the various data type classes.

sphuber:
> This may be a bit ambitious, however, and it would be easier to collaborate on the design of the new data types in the new repository

Fine by me to separate the discussion of changing the design of some of the existing plugins to another place, but I want to reiterate that if we want to change them, we _have_ to do this in conjunction of them moving out and provide a migration in `aiida-core`. This has to be a concerted change and cannot be done later. I know that @azadoks and @ramirezfranciscof were looking into a system of allowing data plugin migrations to be provided by plugins themselves, but I think in working out the first sketches I think they may have come to see the origin of my pessimism of us being able to provide such functionality.

PR #30, opened 2021-12-09
Proposal to replace RabbitMQ

muhrin:
- [X] Used [AEP template](../blob/master/000_aep_guidelines/readme.md) from AEP 0
 - [X] Status is `submitted`
 - [X] Provided github handles for authors

### DIFF
--- a/006_replace_rabbitmq/readme.md
+++ b/006_replace_rabbitmq/readme.md
@@ -1,0 +1,60 @@
+# Remove the dependence on RabbitMQ
+
+| AEP number | 000                                                          |
+|------------|--------------------------------------------------------------|
+| Title      | Remove the dependence on RabbitMQ                            |
+| Authors    | [Martin Uhrin](mailto:martin.uhrin@epfl.ch) (muhrin), [Chris J. Sewell](mailto:christopher.sewell@epfl.ch) (chrisjsewell)|
+| Champions  | [Martin Uhrin](mailto:martin.uhrin@epfl.ch) (muhrin), [Chris J. Sewell](mailto:christopher.sewell@epfl.ch) (chrisjsewell)|
+| Type       | S - Standard Track AEP                                       |
+| Created    | 09-Dec-2021                                                  |
+| Status     | submitted                                                    |
+
+## Background 
+Currently AiiDA uses RabbitMQ as the communication backbone for the daemon.  RabbitMQ is used to queue up tasks, send action messages (asking a worker to pause/play/kill a process) and to send broadcasts informing any listeners of a state change that occurred in a process (e.g. a parent AiiDA process can listen for the termination of its child).
+
+While providing some useful features and functionalities, the use of RabbitMQ comes with a few drawbacks and problems:
+1.  Setting up AiiDA is made more complicated as an OS level service needs to be installed.
+2.  There is no API to introspect RabbitMQ queues to determine what is and what isn't being worked on.
+3.  Desynchronisation between the database state sometimes happens (e.g. a process is still in a non-terminal state but there is no corresponding RabbitMQ message queued, and therefore the process will never continue).
+4.  RabbitMQ is [making changes](https://github.com/rabbitmq/rabbitmq-server/pull/2990) that will see tasks timed out within 30 minutes.  This is not controllable from user space and, what's more, it has been [made clear](https://github.com/rabbitmq/rabbitmq-server/discussions/3345) that long-running tasks are not the intended use case for RabbitMQ.
+
+## Proposed Enhancement 
+This AEP proposes to drop RabbitMQ replacing it by a suitable stand in that does not rely on an external service but retains the core properties (e.g. reliability, responsiveness, scalability, etc).
+
+## Detailed Explanation 
+
+For now this AEP does not propose a concrete alternative but rather fleshes out the requirements which will inform the choice/implementation of a RabbitMQ replacement.
+
+### Original requirements
+
+The design of the current (RMQ based) system was designed to meet the following requirements:
+0. The messaging system should be able to handle three types of communication:
+   1. Task queues (e.g. for a client to submit a process to be run by a worker)
+   2. Remote Procedure Calls (e.g. asking a process to pause/play/kill)
+   3. Broadcasts (e.g. a parent listening for the termination of a child)
+1. An AiiDA process should only ever be ran by one worker at a time
+2. An error can lead to an AiiDA process never running but should not lead to the process being executed by multiple workers simultaneously
+3. The messaging system should be able to handle both long running tasks (days) and many short running tasks (seconds), some of which may be dependent, without unnecessary delays (at most seconds)
+4. The messaging system should not lead to CPU load when nothing is happening (e.g. as a result of frequent polling)
+5. The daemon running processes and the user submitting them should be able to be on separate machines with communication handled by standard protocols (e.g. TCP/IP)
+6. The memory requirements of the system (both RAM and disk) should not grow over time beyond that which is proportionate to the number of processes actively running (i.e. messages referencing terminated or deleted process should be guaranteed to be removed in a timely manner)
+
+
+### Changes to requirements
+
+It is not clear that we need requirement 5.  Currently, it is not possible to run a daemon on a separate machine from the AiiDA database, principally because the repository assumes that the files are stored on the machine where the AiiDA API calls are being made.  This may be possible if an object store backend (e.g. S3) were to be implemented but it's not clear that this will ever be done as the current model of one AiiDA instance per user is prevalent.
+
+In dropping requirement 5 this would open the door to implementing a mechanism that uses the PID to determine if a worker is running or not. 
+
+## Pros and Cons 
+
+Pros and cons of dropping RMQ:
+
+### Pros
+* Installing AiiDA would become simpler, not requiring the installation of an OS service
+* Monitoring which processes are queued or actively running would be possible
+* Depending on the implementation, load-balancing could be greatly simplified (doing away with the need to tune number of worker _and_ slots per worker)
+
+### Cons
+* We may need to develop some in-house code to replace RMQ
+* Any new system will need to be thoroughly tested to ensure that the requirements are met.  This will be challenging for rare failure modes and situations that require large amounts of processes to be submitted/running.


chrisjsewell:
Cheers @muhrin, very helpful, will read through in due course.
Also just to link to the documentation and discussion I've recently had on this topic:

- https://github.com/chrisjsewell/aiida-process-coordinator/blob/main/README.md
- https://github.com/chrisjsewell/aiida-process-coordinator/discussions/4

giovannipizzi:
Thanks! looks good, I'm not sure however if the requirement "The daemon running processes and the user submitting them should be able to be on separate machines with communication handled by standard protocols (e.g. TCP/IP)" is really needed - I would drop it. I don't know if this is what you mention in your AEP, however the numbers are different (you discuss dropping requirement 5 but I see a different number). But from the content, we're discussing the same thing?

giovannipizzi:
(sorry, closed by accident)

006_replace_rabbitmq/readme.md
```diff
@@ -0,0 +1,60 @@
...
+5. The daemon running processes and the user submitting them should be able to be on separate machines with communication handled by standard protocols (e.g. TCP/IP)
+6. The memory requirements of the system (both RAM and disk) should not grow over time beyond that which is proportionate to the number of processes actively running (i.e. messages referencing terminated or deleted process should be guaranteed to be removed in a timely manner)
+
+
+### Changes to requirements
+
+It is not clear that we need requirement 5.  Currently, it is not possible to run a daemon on a separate machine from the AiiDA database, principally because the repository assumes that the files are stored on the machine where the AiiDA API calls are being made.  This may be possible if an object store backend (e.g. S3) were to be implemented but it's not clear that this will ever be done as the current model of one AiiDA instance per user is prevalent.
+
+In dropping requirement 5 this would open the door to implementing a mechanism that uses the PID to determine if a worker is running or not.
```
greschd, line 47: > this would open the door to implementing a mechanism that uses the PID to determine if a worker is running or not.

While I don't have a qualified opinion on the requirement, I don't see how dropping it is necessary to achieve this. If the "communication backbone" lives on the "daemon running processes" side of things, it should both be possible to identify workers by PID, and remotely control it from a "user client".

Am I missing something obvious here?

muhrin: Ah, yes, you're right!  I confused myself and now realise this is only an issue if you (somehow) want to support the worker and daemon being on different machine which I realise now is never going to be a use case for AiiDA so I think we're all good.  I'll update this tomorrow.

muhrin:
Thanks @giovannipizzi , you're quite right.  This point isn't explained very well, I'll rewrite it tomorrow to explain it starting from the use case I had in mind many years ago when we started this process.  Basically I would like for it to be possible to have, say, a group AiiDA instance with many users concurrently interacting with the same AiiDA DB.

006_replace_rabbitmq/readme.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Original requirements
+
+The design of the current (RMQ based) system was designed to meet the following requirements:
+0. The messaging system should be able to handle three types of communication:
+   1. Task queues (e.g. for a client to submit a process to be run by a worker)
+   2. Remote Procedure Calls (e.g. asking a process to pause/play/kill)
+   3. Broadcasts (e.g. a parent listening for the termination of a child)
+1. An AiiDA process should only ever be ran by one worker at a time
+2. An error can lead to an AiiDA process never running but should not lead to the process being executed by multiple workers simultaneously
```
chrisjsewell, line 36: This seems very lenient no?
For sure, it is a problem more than one people have encountered with the current system; rabbitmq loses the task and the `ProcessNode` now is in an unterminated state, with a checkpoint, but nothing trying to run it, and no way to know if rabbitmq still has the tak or not.
Then, if you gamble that there is indeed no current task, you have to run some API "hacks" to create a submit a new `continue` task.

Would it then be a change of requirements, that there should indeed be a guarantee that processes are run to termination?
Or at least that the task queue is easily introspectable, and that there is a front-end mechanism in aiida to start a new `continue` task on an unterminated `ProcessNode` that no longer has one

sphuber: I haven't had the time to go through the entire proposal yet, but I will just say that I agree with this observation. Not being able to easily introspect the state of the task queue has been a great source of problems. It made it very difficult to debug, even when you had access to the machine. At the very least this should be made very clear, if we cannot guarantee it to not occur of course, that would always be best.

muhrin: @chrisjsewell , yes, this isn't very clear.  This point was meant to address a binary decision that must be made by any concurrent system.  In such systems one can only _guarantee_ one or other of the following:
1. A task will run at least once, but possibly more times.  This is the most appropriate choice for systems that have tasks that are idempotent and where the resources consumed by a task not a concern.  Clearly, this choice is not a good fit for AiiDA.
2. A task will run zero or one time.  This is pretty much the only choice that makes sense for AiiDA hence the wording of requirement 2.

All this said, this requirement wasn't meant to imply that we shouldn't do everything in our power to make sure that the probability that a task runs zero times is low (and in fact, this is exactly one of the things that RMQ is _meant_ to do).

As for making it easy to resubmit a task that failed to run at all, I think we all agree that this should be a high priority.

006_replace_rabbitmq/readme.md
muhrin: Ok, this sentence has been dropped as it didn't make sense.

006_replace_rabbitmq/readme.md
muhrin: Added a clarification to the SI of the document.

006_replace_rabbitmq/readme.md
```diff
@@ -0,0 +1,106 @@
...
+4. The messaging system should be able to handle both long running tasks (days) and many short running tasks (seconds), some of which may be dependent, without unnecessary delays (at most seconds) ([elaboration](#requirement-4))
+5. The messaging system should not lead to CPU load when nothing is happening (e.g. as a result of frequent polling)
+6. The daemon running processes and the user submitting them should be able to be on separate machines with communication handled by standard protocols (e.g. TCP/IP) ([elaboration](#requirement-6))
+7. The memory requirements of the system (both RAM and disk) should not grow over time beyond that which is proportionate to the number of processes actively running (i.e. messages referencing terminated or deleted process should be guaranteed to be removed in a timely manner)
+8. It should be possible to dynamically scale the number of workers to accommodate an increase or decrease in the load on the daemon and for the user to control the permitted load on their machine.
+
+
+### Proposed replacement
+
+This section will evolve into an actionable solution but, for now, will be used to flesh out and discuss a solution.
```
csadorf, line 47: I think for a balanced and more complete discussion, we should explicitly consider alternative solutions in addition to Chris' pure-Python prototype. I understand that needing this service comes with a lot of issues in itself, as outlined in the introduction, but I would like to say with confidence that the problems to be addressed with the removal of RabbitMQ could not also be addressed with

1. Using RabbitMQ differently. Is it an absolute requirement that RabbitMQ manages the complete state of the process over its lifetime? Or could we also just push events like the "request to start a process", "a change in process state", and so on and then have more lightweight process handlers react to that?
2. Would a switch to something like [Apache Kafka](https://kafka.apache.org/) (or a pure-Python alternative like [faust-streaming](https://github.com/faust-streaming)) be a suitable alternative?

Especially using something like Kafka would potentially give us some avenues for network-distributed and multi-user support. It is possible (maybe even probable) that those are not the right models, but I think we should make sure that we can say that with confidence before venturing on this major architecture change.

csadorf (changes requested):

006_replace_rabbitmq/readme.md
sphuber: Think you raise a valid point @csadorf that this change should certainly not be taken too lightly, and so merits to have other potential solutions investigated.

Just wanted to already quickly give some feedback on your point 1. I think we currently already use RabbitMQ in as light a manner as possible. What happens is exactly as you describe: when you submit a process, a request to do so is sent to RabbitMQ, which will then send this "task" to a worker. As part of its guarantees that each task will be at some point executed, RabbitMQ will wait for the worker to confirm the task is done and if not, it will at some point send it to another worker, if the heartbeat is missed (and so it has to assume the worker died). This is where the problem lies though as RabbitMQ was designed for short running tasks (order of seconds/minutes) and not the tasks typically run with AiiDA (hours/days). This can be configured, but not out of the box, which is the reason that we are considering replacing it. I don't think that we can get around this feature of RabbitMQ though as the whole reason we selected it in the first place was for its guarantees of eventually completing the tasks it receives and its persistence.

sphuber: At least for that part, but we also use the message broker part. As a process changes state, we emit broadcast. This is very useful as it makes the system event-based instead of polling based. For example, parent processes that are waiting for their child process to finish will be notified immediately when this happens and can continue running. With a polling system you always have a delay. And making this delay too short tends to overload the system unnecessarily.

csadorf: I am not even remotely convinced that this is the right solution, but trying to understand: in principle we could push a request for a process to be executed onto a Kafka topic (event log), which could then be picked up by a worker. The same topic could then also be used to communicate back any state changes.

muhrin: I do like the sound of the Kafka architecture but conceptually for this use case we wouldn't really be using it in its most powerful mode (an event log).  I did once, mostly jokingly, propose rewriting the AiiDA DAG in Kafka which would be a good fit for an event store.

Anyway, to the matter at hand, I think Kafka _could_ work but I do think there is a pretty large advantage to _not_ requiring the installation of a system level service (which I think Kafla is).  I don't know if faust-streaming would get around this.

Essentially, whatever solution we adopt there will need to be at least two components:
1) A mechanism to handle incoming messages (through polling or events), and,
2) A persistent store of messages, ideally with some guarantees.

I played around last year with using `xmlrpc` (a standard python module) for the event based messages part (it was pretty easy to create a kiwipy compatible communicator), but this would still require the storage component which in Chris' proposal comes in the form of additional database tables.  Personally I'm not so keen for AiiDA to take responsibility for the storage of events but then I haven't seen any persistent, non-OS-service based solutions.

chrisjsewell: >  to not requiring the installation of a system level service 

Yeh I'm very keen to avoid swapping out one "user complication" for another 😬 
I'll  have alook at faust though

Just to note here, I'm going to be delving in to the internals of rabbitmq a bit more in the coming month(s), to understand it more thoroughly.
Crudely speaking though, its not magic; it is a daemon service, that uses a database to persist data, and a JSON based communication protocol (with additional caching features etc https://blog.rabbitmq.com/posts/2015/10/new-credit-flow-settings-on-rabbitmq-3-5-5)
With the aiida daemon, we already have a daemon service, that uses a database to persist data, plus we also have direct control over what process are running etc, which rabbitmq does not, and I feel we are just duplicating effort to some extent

chrisjsewell: Here's a nice summary: https://hvalls.dev/posts/polling-long-polling-push, which specifically references:

- RabbitMQ uses a push model, which is appropriate when you have limited clients/consumers and messages
- Kafka is an example of technology implementing a long polling model

csadorf: > Essentially, whatever solution we adopt there will need to be at least two components:
> 
> A mechanism to handle incoming messages (through polling or events), and,
> A persistent store of messages, ideally with some guarantees.
> 

Unless I completely misunderstand the scope and mission of Kafka, that is its **exact** purpose. So it sounds to me like https://github.com/faust-streaming might indeed be important to evaluate then. Not only with respect to its fit as a technical solution, but also with respect to its sustainability (the project was abandoned by its original creator "Robinhood" and is currently maintained as a community fork).

unkcpz:
Hi all,
I open a new AEP for how I think it can be implemented at https://github.com/aiidateam/AEP/pull/42, please have a look and give your feedback.

PR #31, opened 2021-12-10
✨ NEW: AEP For push/pull mechanism

chrisjsewell:
Empty at the moment!
But I will write up the notes from our discussion here, then start to flesh it out

 - [ ] Status is `submitted`

### DIFF
--- a/000x-push-pull.md
+++ b/000x-push-pull.md
@@ -1,0 +1,3 @@
+# AEP for Push/Pull mechanism
+
+TODO ...


chrisjsewell:
Some initial notes:

# Archive Push-Pull Mechanism

## Introduction

AiiDA provides an archiving feature, for exporting/importing provenance graphs to a self-contained file. With AiiDA 2.0, the archive format and export/import API has been re-designed, allowing for the possibility of incremental updates to an archive and selective imports. This now opens the door for a push/pull feature, whereby collaborators could share a central "provenance repository", push/pulling delta differences in much the same way as collaborators use GitHub remotes and local git checkouts

https://github.com/aiidateam/aiida-core/issues/4535 

## Discussion

Partial pulls

Hash of uuids

Security

Checksums

Enforce schema version of backend (could be a pain for users)
Think of changes with sharing in mind

Low-level vs high-level migrations

Pulling:

Create an empty profile with connection to remote
Ask remote for node UUID(s)
Handshake
Remote gives back all entity UUIDs required for the node’s provenance
Fetching …
Must be allowed for connection to be lost and we still have the data to that point stored
Temporary profile? 
Copying node topology (links)


Make AEP

Always push

PR #33, opened 2021-12-16
✨ NEW: Add RTD site

chrisjsewell:
Because why not!

 - [ ] Status is `submitted`

### DIFF
--- a/000_aep_guidelines/readme.md
+++ b/000_aep_guidelines/readme.md
@@ -74,7 +74,7 @@ A summary for pros and cons for the proposed enhancement, detailing e.g. which p
    and put readme file in [Markdown](https://guides.github.com/features/mastering-markdown/) format inside this folder.
      * You may use this `readme.md` as a template.
      * You may use the folder to include additional files relevant to your enhancement proposal.
-     * Add your AEP to the `README.md` at the top level of the repository.
+     * Add your AEP to the `README.md` at the top level of the repository, and in to `_toc.yml`.
      * Consider discussing your proposal informally with an [AiiDA team](https://github.com/orgs/aiidateam/people) member to get an initial reaction
        and, potentially, another champion for your proposal.
 3. Commit your changes and submit a [pull request](https://help.github.com/en/articles/creating-a-pull-request) to the AEP repository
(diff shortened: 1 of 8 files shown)


sphuber:
Nice. Should we maybe put the doc-specific files in a `docs` sub directory to keep root from getting to cluttered?

chrisjsewell:
> Should we maybe put the doc-specific files in a docs sub directory to keep root from getting to cluttered?

Ah but then I believe you would have to put all the documentation under the `docs` directory, which is obviously everything in the repository lol

chrisjsewell:
There you go: https://aep--33.org.readthedocs.build/en/33/

chrisjsewell:
All done on my end.
Note, when you add a new AEP, you just need to add it to `_toc.yml`

sphuber:
> Note, when you add a new AEP, you just need to add it to `_toc.yml`

Would be good to add this to the steps for submitting an AEP in AEP 000.

_toc.yml
```diff
@@ -0,0 +1,11 @@
+root: README.md
+subtrees:
+- caption: Proposals 
+  entries:
```
ltalirz, line 4: There's no way to automate this, right?

000_aep_guidelines/readme.md
```diff
@@ -74,7 +74,7 @@ A summary for pros and cons for the proposed enhancement, detailing e.g. which p
    and put readme file in [Markdown](https://guides.github.com/features/mastering-markdown/) format inside this folder.
      * You may use this `readme.md` as a template.
      * You may use the folder to include additional files relevant to your enhancement proposal.
-     * Add your AEP to the `README.md` at the top level of the repository.
+     * Add your AEP to the `README.md` at the top level of the repository, and in to `_toc.yml`.
```
chrisjsewell, line 77: is that ok?

ltalirz (approved):
Thanks @chrisjsewell !

>> Should we maybe put the doc-specific files in a docs sub directory to keep root from getting to cluttered?

> Ah but then I believe you would have to put all the documentation under the docs directory, which is obviously everything in the repository lol

I would also have preferred that but if there's no way to do a `../` there, it's ok with me.

000_aep_guidelines/readme.md
sphuber:
```suggestion
     * Add your AEP to the `README.md` and `_toc.yml` files at the root of the repository.
```

sphuber (approved):

_toc.yml
chrisjsewell: nope, not currently

PR #37, opened 2023-01-22
Code computer setup

unkcpz:
 - [x] Status is `draft`

### DIFF
--- a/008_easy_to_use_code_registry/readme.md
+++ b/008_easy_to_use_code_registry/readme.md
@@ -1,0 +1,47 @@
+# AEP 000: AiiDA Enhancement Proposal (AEP) Guidelines
+
+| AEP number | 008                                                          |
+|------------|--------------------------------------------------------------|
+| Title      | New aiida-code-registry for easy code/computer setup                  |
+| Authors    | [Jusong Yu](mailto:jusong.yu@psi.ch) (unkcpz) |
+| Champions  | [Jusong Yu](mailto:jusong.yu@psi.ch) (unkcpz) |
+| Type       | P - Process                                                  |
+| Created    | 08-Jan-2023                                                  |
+| Status     | draft                                                  |
+
+## Background 
+
+- computer/code setup is not convinient in AiiDA
+- we have aiida-code-registry for the purpose that user can store the config files for future setup and for sharing
+
+## Problems of current aiida-core-registry
+
+- aiidalab-widgets-base using aiida-code-registry as database provide widget to setup computer/code but lack of flexibility for very detail parameters.
+- lack of capability to export the config file to store for future setup and for sharing purpose.
+
+## Goals and proposed Enhancement
+
+- new subcommand `bytemplate` for `verdi code create bytemplate <entrypoint>` which read from j2 template and dynamically create options ask for inputs and create config file and pass to `verdi code create <entrypoint>`.
+- Same concept can be used for computer setup and computer configuration as well. To induce the complexity, it better to have `verdi computer create` to replace `verdi computer setup` as well.
+
+- In aiida-code-registry, users are encourage to upload the template for computer/code setup. 
+- It also good to have a well designed web page that list all the template config files and user can search for the config files they need. It need to easy to copy and better to generate a url with only the raw template data that can directly pass to command for setup.
+
+- verdi command to export the config files. Attampts are https://github.com/aiidateam/aiida-core/pull/5860 and https://github.com/aiidateam/aiida-core/pull/3616
+- not only the config of one code/computer setup but all computer/code configs of a profile can be exported into a particular format that can be store in the aiida config folder. It can be then shared and for easy load the configs. The idea is bring up by ltailz in https://github.com/aiidateam/aiida-code-registry/pull/62.
+
+## Roadmap of the implamentation
+
+
+
+## Pros and Cons 
+
+### Pros
+* The public discussion of enhancement proposals lets the entire AiiDA community see what developments lie ahead
+  and allows those interested to actively participate in shaping them
+* AEPs provide basic guidance on how to "make a case" for an enhancement such that it can be seriously discussed
+  before investing efforts in its implementation
+* The corresponding pull requests provide a public record of the decision process in case questions arise later
+
+### Cons
+* Introduce new command for code/computer setup and coexist with old setup command may bring complicity to new user and not familiar for old user.
--- a/008_easy_to_use_code_registry/readme.md
+++ b/008_easy_to_use_code_registry/readme.md
@@ -1,4 +1,4 @@
-# AEP 000: AiiDA Enhancement Proposal (AEP) Guidelines
+# AEP 008: AiiDA code registry for easy setup of computer/code
 
 | AEP number | 008                                                          |
 |------------|--------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


unkcpz:
I am not sure about the implementation details for all tasks mentioned in the AEP, it is the overview of the issues and things that need to be fixed and potential solutions.

008_easy_to_use_code_registry/readme.md
```diff
@@ -0,0 +1,109 @@
+# AEP 008: AiiDA code registry for easy setup of computer/code
+
+| AEP number | 008                                                          |
+|------------|--------------------------------------------------------------|
+| Title      | New aiida-code-registry for easy code/computer setup                  |
+| Authors    | [Jusong Yu](mailto:jusong.yu@psi.ch) (unkcpz) |
+| Champions  | [Jusong Yu](mailto:jusong.yu@psi.ch) (unkcpz) |
+| Type       | P - Process                                                  |
```
ltalirz, line 8: I think this is a "S - Standard Track"  AEP

008_easy_to_use_code_registry/readme.md
```diff
@@ -0,0 +1,109 @@
...
+- Design the web page to parse and show the YAML/j2 config file back in the new code registry page.
+
+## Pros and Cons 
+
+### Pros
+* The public discussion of enhancement proposals lets the entire AiiDA community see what developments lie ahead
+  and allows those interested to actively participate in shaping them
+* AEPs provide basic guidance on how to "make a case" for an enhancement such that it can be seriously discussed
+  before investing efforts in its implementation
+* The corresponding pull requests provide a public record of the decision process in case questions arise later
```
ltalirz, line 106: these are prose of the AEP process rather than of your proposal

008_easy_to_use_code_registry/readme.md
```diff
@@ -0,0 +1,109 @@
...
+If for each domain, there is one file for all setup, we need provide a way to parse these information back to template or YAML file so can be used.
+A new and well designed code registry page is needed to make this possible.
+It should looks the same as [AiiDA plugin registry](https://aiidateam.github.io/aiida-registry/) with extra functionalities that the config file can be parsed back from information given and can be a pure YAML/j2 page having a specific URL to be downloaded or for setup the code directly.
+It can be very useful that the command for setup computer/code directly show in this page (toggle to hide/show) and user can copy and paste to setup.
+
+- It also good to have a well designed web page that list all the template config files and user can search for the config files they need. It need to easy to copy and better to generate a url with only the raw template data that can directly pass to command for setup.
+
+## Roadmap of the implamentation
+
+- The first step would be to make it able to export the setup from database for both computer and code with the data stored in the database and for the computer configure data.
```
ltalirz, line 94: do you have a specific use case for this?

this is nice to have but, at least for me, the important part is the templating functionality of the registry (and I would therefore focus on that)

008_easy_to_use_code_registry/readme.md
```diff
@@ -0,0 +1,109 @@
...
+- Then I'll implement setup from template file.
```
ltalirz, line 95: What I would like to see in this AEP is a concrete proposal for the templating mechanism.

I've been using my proposal https://github.com/aiidateam/aiida-code-registry/pull/62/files internally already for a while and found it quite useful, but I am open to discussion.

One concrete point I would like to open up for discussion is the question whether we should continue forcing users to provide the absolute path to an executable.
In my use cases, this often leads to duplication of information, since in addition to a `module load quantum-espresso/...` in the `prepend_text`, I also need to provide the absolute path to the executable. This path can change, for example, depending on where the corresponding directory is mounted. It also makes it impossible to rely on auto-detection of, say, the microarchitecture at the `module load` level (e.g. by pointing the module path to different locations depending on the microarchitecture, so that `module load quantum-espresso/7.1` will  automatically load different modules on an AMD zen3 node than on an Intel skylake node).

Recording the absolute path of the executable in the code does provide additional provenance, but in my view users should be able to decide by themselves whether they want to provide it or not.

P.S. For containerized codes, the absolute path actually no longer provides additional provenance, so there the requirement can be skipped.

008_easy_to_use_code_registry/readme.md
unkcpz: Yes, changed.

unkcpz:
@ltalirz thanks for the review and advice! I make it a draft, and for sure it needs more effort to polish and add more details.

008_easy_to_use_code_registry/readme.md
unkcpz: I don't have a specific use case, I simply write it since it is a low-hanging task and will benefit usability a lot. You are correct, it is not a crucial part of this AEP and I can imagine at the end this may even not a useful one if only the current YAML file is exported but we will then support the template and directly set it from the aiida code registry.

008_easy_to_use_code_registry/readme.md
unkcpz: I think your PR contains all components for the templating config. One thing I want to add in the implementation is to move it to aiida-core to make it possible to setup code/computer with verdi. 
I also do not fully agree with either the current many YAML files for computers/codes, or in your implementation all in one big file. I prefer in the middle that to organize every domain has a file to define all the codes/computers, and having a web API take care of converting it to the config files that can be use in `verdi` command.

unkcpz: > What I would like to see in this AEP is a concrete implementation proposal for the templating mechanism and user interface.

It is my fault that I don't have a very clear vision of how to get it done, I made some attempts to borrow your PR and try to integrate it into the verdi using `DynamicEntryPointCommandGroup`. But no luck to make it fully work as I expected.

sphuber: > Recording the absolute path of the executable in the code _does_ provide additional provenance information, but in my view users should be able to decide by themselves whether they want to provide it or not.

I discussed this current limitation at the MARVEL retreat with @chrisjsewell and we agreed that it should probably be removed. This was added before I joined so I can't vouch for the original motivation, but I guess the reason was indeed to increase the provenance somewhat. If the path is relative, a module load in the `prepend_text` could effectively change the version that is being executed. However, it is not like this is impossible with an absolute path, so it merely gives a false sense of provenance. So I am all for removing the limitation

008_easy_to_use_code_registry/readme.md
```diff
@@ -0,0 +1,104 @@
...
+| Title      | New aiida-code-registry for easy code/computer setup                  |
+| Authors    | [Jusong Yu](mailto:jusong.yu@psi.ch) (unkcpz) |
+| Champions  | [Jusong Yu](mailto:jusong.yu@psi.ch) (unkcpz) |
+| Type       | S - Standard Track                                                |
+| Created    | 08-Jan-2023                                                  |
+| Status     | draft                                                  |
+
+## Background 
+
+For the beginners and even for the experienced AiiDA user, setting up computers and codes is still a tidious mission.
```
chrisjsewell, line 14:
```suggestion
For the beginners and even for the experienced AiiDA user, setting up computers and codes is still a tedious mission.
```

(if you haven't already, I would suggest installing https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker 😉 )

008_easy_to_use_code_registry/readme.md
chrisjsewell: Yep seconded

008_easy_to_use_code_registry/readme.md
unkcpz: Nice, thanks! Actually, I edit in the github web interface and get spell checked by Grammerly, which is super inconvenient 😆.

008_easy_to_use_code_registry/readme.md
ltalirz: > One thing I want to add in the implementation is to move it to aiida-core to make it possible to setup code/computer with verdi.

That is fine but could always be done in a second step - i.e.:
(1) have a simple cli that ships with the registry; 
(2) once people are happy with it, integrate it into `verdi`

> I also do not fully agree with either the current many YAML files for computers/codes, or in your implementation all in one big file. I prefer in the middle that to organize every domain has a file to define all the codes/computers

There is perhaps a misunderstanding: my implementation is fully compatible with what you suggest, and I agree that there should be one file per domain.

PR #39, opened 2023-07-16
AEP: new docker stack to handle services

unkcpz:
 - [x] Status is `draft`

### DIFF
--- a/009_improved_docker_images/readme.md
+++ b/009_improved_docker_images/readme.md
@@ -1,0 +1,163 @@
+# AEP 009: Improved Docker images
+
+| AEP number | 009                                                             |
+|------------|-----------------------------------------------------------------|
+| Title      | Improved Docker images                                          |
+| Authors    | [Jusong Yu](mailto:jusong.yeu@gmail.com) (unkcpz)               |
+| Champions  | [Jusong Yu](mailto:jusong.yeu@gmail.com) (unkcpz)               |
+| Type       | S - Standard                                                    |
+| Created    | 16-July-2023                                                    |
+| Status     | implemented                                                     |
+
+
+## Background and problem description
+
+AiiDA provides and maintains a Docker image to offer a clean and isolated environment that is easy to setup for users and developers.
+However, the the current Docker stack has a number of problems and limitations:
+- The `aiidateam/aiida-prerequisites` image, on which it is based, is not maintained and updated regularly.
+  (Partly because the `aiidalab-docker-stack` replaced `aiida-prerequisites` with `jupyter/minimal-notebook` as its base image.)
+- The image is not able to run on `linux/arm64` architecture, therefore not able to run on Apple M1 chip which is now widely is use by the community.
+- The RabbitMQ and PostgreSQL services are not properly installed and configured inside the container, causing the user to have to start them manually.
+As a result, it is not straightforward to create a container and start using AiiDA, as evidenced by the [length of the documentation](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/run_docker.html#intro-get-started-docker).
+
+One of the main reasons for providing a Docker image with `aiida-core` was to provide a base image for the `aiidalab-docker-stack` to build on.
+However, thanks to the fantastic work by @csadorf, the Docker stack of the `aiidalab` organization is now built on top of `jupyter/minimal-notebook` instead of `aiidateam/aiida-core`.
+The [`aiidalab-docker-stack`](https://github.com/aiidalab/aiidalab-docker-stack) comes in four variants:
+
+- `base` – A minimal image that comes with AiiDA pre-installed and an AiiDA profile set up.
+- `base-with-services` – Like `base`, but the services required by AiiDA (PostgreSQL and RabbitMQ) are installed in the container and automatically launched on startup.
+- `lab` – Like `base`, but uses the AiiDAlab home app as the primary interface (the standard JupyterLab interface is also available).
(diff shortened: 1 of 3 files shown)


unkcpz:
https://aep--39.org.readthedocs.build/en/39/009_new_docker_stack_image/readme.html

sphuber:
Thanks @unkcpz . I have read through the AEP and most of it seems good to me. I have added a commit with some small fixes to readability. This was easier than doing it through suggestions through the PR review interface. Feel free to point out changes that are incorrect so we can fix them.

I have one major question I think: I am bit confused on the image names. It seems that the docker files in `aiida-core` are called `base` and `base-with-service`. But the names on docker hub are supposed to be `base` and `aiida-core`? Why are they different? And would `base` and `full` be simpler and convey just as much? And do we need to prefix them with `aiida-` to make clear that they are images for AiiDA? E.g., `aiida-base` and `aiida-full`. Or is the context of the image always clear?

009_new_docker_stack_image/readme.md
```diff
@@ -10,111 +10,104 @@
...
 - `full-stack` – Our most comprehensive image, like `lab`, but also comes with services pre-installed and launched.
 
-With this new way of maintaining aiida/aiidalab images, the `aiida-prerequisites` is not used anymore. We need to find a way to elegantly replace the old approach and make the new approach more automatic and easy to maintain.
-
-The `base` and `base-with-services` images are the two images that are contains only `aiida-core` but has following problems:
-
-- Useless to use `jupyter/minimal-notebook` as base image
-
-The `jupyter/minimal-notebook` will mapping the port and start a jupyter notebook as the main service of the container which for the `aiida-core` user it is useless.
+Due to the switch to `jupyter/minimal-notebook` as the base image, the `aiida-prerequisites` image has become superfluous and they no longer rely on the `aiidateam/aiida-core` image.
```
unkcpz, line 32: It was `aiidateam/aiida-core` image rely on the `aiida-prerequisites`

009_new_docker_stack_image/readme.md
```diff
@@ -10,111 +10,104 @@
...
 
-3. Images support both `linux/amd64` and `linux/arm64`, and include basic system tools: `vim`, `git`, `conda`/`mamba` for quick development environment buildup. 
+Two new Docker images will be maintained within the `aiida-core` repository and pushed to the `aiidateam` docker/ghcr hub namespace:
 
-4. docker run command will start the container and provide a ready to use AiiDA environment for users to use.
+- `aiidateam/base`: Image contains just `aiida-core` installed, which can be used with independent RabbitMQ & PostgreSQL services, e.g., through `docker-compose`.
+- `aiiidateam/aiida-core`: Image built on top of `base` that inclues RabbitMQ & PostgreSQL services.
 
-5. Hide `base` and `base-with-services` images from docker hub and ghcr.
+### Hide `base` and `base-with-services` images from docker hub and ghcr
```
unkcpz, line 70: Sorry that I confused you her. This is a 5th point and the goal was to not show `base` and `base-with-services` images for aiidalab. In the end, we will have four official maintained images:
- `aiidateam/base`
- `aiidateam/aiida-core`
- `aiidalab/lab`
- `aiidalab/full-stack`

unkcpz (commented):
@sphuber thanks a lot for the correction, two places may change what I want to express initially.

> It seems that the docker files in aiida-core are called base and base-with-service. But the names on the docker hub are supposed to be base and aiida-core?

I kept naming the image in `ghcr` and `dockerhub` the `aiida-core` since I thought it was more natural for people who already use the `aiidateam/aiida-core` image, no?

The `base-with-services` is directly from aiidalab, I think the name explains what it is. The `full` seems a bit too far, what if we want to have an image that includes all the developing tools installed in the image? I have no preference for adding `aiida-` in front. Maybe not be necessary since for the image we already own the `aiidateam` and `aiidalab` namespace.

I see one downside of using name without `aiida-` that is when the image is used as base image by other people and upload to dockerhub, if they keep using `base` would make no sense. So what about we name these two `aiidateam/aiida-base` and `aiidateam/aiida-core` (If the first point I mention is not important, `aiida-full` is also an option)?

009_new_docker_stack_image/readme.md
sphuber: It is indeed worded a bit unclear. With the "they no longer rely" I meant the "they" to refer to the `aiidalab` organization, which now just uses `jupyter/minimal-notebook`. Will improve the language a bit

sphuber:
> I see one downside of using name without `aiida-` that is when the image is used as a base image by other people and upload to dockerhub, if they keep using `base` would make no sense. So what about we name these two `aiidateam/aiida-base` and `aiidateam/aiida-core`

I think we should definitely prefix with `aiida-` for the reasons I explained in another comment but yours is also important (what happens if the image is taken out of the `aiidateam` namespace, `base` is too generic). We can bring it up during the team meeting today. I feel `aiida-base` and `aiida-core` are too close and to me it is not clear what the difference would be. Maybe `aiida-core-base` and `aiida-core-services` ? Since the difference is that one image is a base image and the other one includes the services.

sphuber:
After discussion during the team meeting, we settled on `aiida-core` and `aiida-core-with-services` for the names of the two images.

@unkcpz I will update the AEP, can you take care of updating the PR on `aiida-core`? Then I will take care of merging this

unkcpz:
> I will update the AEP, can you take care of updating the PR on aiida-core? Then I will take care of merging this

Sure, I'll update the PR. Thanks for taking care of AEP.

sphuber (approved):
Thanks a lot @unkcpz great work!

PR #44, opened 2024-11-29
WorkGraph

superstar54:
 - [x] Status is `draft`


010_workgraph/readme.md
```diff
@@ -0,0 +1,278 @@
...
+```
+
+**Linking Tasks:**
+Add task dependencies by connecting the sockets of tasks.
+```python
+wg.add_link(wg.tasks["add"].outputs["result"], wg.tasks["multiply"].inputs["x"])
+```
+
+### Engine
+The `WorkGraphEngine` class is responsible for executing the workflow. It is designed to handle the following features:
```
unkcpz, line 103: The "Engine" as suffix is confusing. For WorkChain there is no need to have a "WorkChainEngine". I'd wondering is layer really necessary? On the other hand, the "engine" can be confused for aiida-core engine module.

superstar54: You can think a `WorkChain` is an engine for the workflow defined in its `outline`.  For each workflow (outline), one need to create a `WorkChain` to execute it. That's the inconvenient part of the WorkChain.
In WorkGraph, we only need one `engine` for all workflows. The workflow can be defined either in the WorkGraph Python API, or from a yml file, or from the GUI.
`WorkGraphEngine` belongs to the `engine` module, like `aiida.engine.processes.workgraph:WorkGraphEngine`.

PR #46, opened 2025-06-19
AEP 010: Event‑based, multi‑instance process scheduler

superstar54:
 - [x] Status is `draft`


010_aiida_scheduler/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+* A **Scheduler list** (DataGrid) with start/stop buttons and live counts.
+
+![Figure 1 – GUI scheduler list](aiida-scheduler-gui-list.png).
+
+* A **Detail view** with editable limits and a real‑time line chart of running vs waiting processes.
+
+![Figure 2 – GUI detail view](aiida-scheduler-gui-detail.png).
+
+* A **Process table** showing individual jobs with play/pause/kill & priority editing.
```
unkcpz, line 135: Please.. don't mix the frontend implementation and backend implemenation, AGAIN.

010_aiida_scheduler/readme.md
```diff
@@ -0,0 +1,198 @@
...
+## Detailed Design
+
+Figure 1 illustrates the interplay between user submission, RabbitMQ, the new Scheduler daemon, AiiDA daemon workers, and the database.
+![Figure 2 – AiiDA process execution flow with Scheduler](aiida-scheduler.png).
+
+---
+
+## Implementation Reference
+
+A proof‑of‑concept lives in **aiida‑workgraph**:
```
unkcpz, line 149: Can this independent of workgraph?

010_aiida_scheduler/readme.md
```diff
@@ -0,0 +1,198 @@
...
+  | `waiting_process`                                | FIFO (First-In, First-Out) list of processes waiting for capacity             |
+  | `running_process`                                | Currently running processes, capped by limits           |
+  | `next_priority`                                  | Decrementing counter so lower numbers ⇒ higher priority |
+  | `max_processes`, `max_calcjobs`, `max_workflows` | User‑defined limits                                     |
+
+* On capacity, pop the highest‑priority process from `waiting_process` and invoke `ProcessController.continue_process(pk)`.
+
+* Detect process completion via RabbitMQ broadcasts (with database polling fallback) and update state.
+
+### Daemonization via Circus
```
unkcpz, line 79: I have a major concern to use circus, not only because it is almost deprecated (I think we are the only maintainer of it and we are not familiar with the core code base of circus), but also it means two type of "daemon" are required.

010_aiida_scheduler/readme.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Proposed Enhancement
+
+We introduce an **event‑driven scheduler** implemented as a lightweight daemon that consumes `CONTINUE_PROCESS` tasks from RabbitMQ, honours user‑defined limits & priorities, and persists its state in the AiiDA database.
+Multiple scheduler instances can run concurrently.
+
+### Naming and Multi‑Instance Support
+
+* Each scheduler is identified by a **user‑supplied name** (e.g. `eiger`, `merlin7-cpu`, `merlin7-gpu`). A common practice is to name it after an HPC *computer* to indicate the set of processes it is meant to *rate‑limit*, but this is **purely conventional**—it does **not** override the `Computer` *input* of the process itself.
+* Users may create **any number of schedulers** to enforce different concurrency / priority policies for different groups of processes.
```
unkcpz, line 59: Be aware that every scheduler need one python process to manage.

unkcpz (changes requested):
I personally would against this proposal. 

1. rmq is already overkill for the regular aiida-core cases, here it is overkill to be used for the "task scheduler". We try to remove rmq, this just makes it much more rely on RMQ. 
2. it brings another type of daemon, then there is a resources manage problem with two type of daemon, which can be dropped first? are they dependent on each other?
3. It depends on `circus` which is deprecated in the python community, we want to get rid of in aiida this proposal just make it more strong depend on it.

010_aiida_scheduler/readme.md
superstar54: I don't understand your question. Could you explain what you mean?

010_aiida_scheduler/readme.md
superstar54: Yes, it is independent of workgraph. Once we decide on a proper name, we can move it out as a separate package.

superstar54:
Hi @unkcpz , thanks for reviewing this proposal!

**Regarding RMQ and Circus**:
I don’t think their usage is a problem. The current design is fully based on AiiDA’s existing implementation, using RMQ and Circus, with no new technologies introduced. This isn’t a downside; in fact, it’s a strong advantage. It allows us to implement the new feature without needing to refactor the core, saving significant effort and maintaining consistency with the current stack.

If, in the future, the stack is replaced (e.g., replacing the daemon manager or message broker), we will update the scheduler in the same way. So there’s no additional effort required beyond what will already be necessary.

Relying on a package in more than one place doesn’t increase the dependency. We're not introducing a new dependency, just using the existing ones more effectively.


---
> **Two types of daemon—can they be dropped independently? Are they dependent?**

They are completely independent. You can stop either one without affecting the other. While this does introduce the need for users to manage an additional scheduler daemon, it's worth noting that, in the case of the `aiida-submission-controller`, users also have to manage a running Python script themselves. In comparison, the proposed scheduler daemon provides a more consistent and user-friendly way to manage this process within the existing AiiDA infrastructure.



Looking at your three main points, I don’t see a strong argument against the proposal. The design works within the current architecture, avoids unnecessary refactoring of the core, and doesn't introduce new complexity in terms of dependencies or system management.

superstar54:
> don't leave the problem to the "future".

Note, it's your arguments that are based on the "future" assumption (which may or may not happen), and this proposal is based on the current AiiDA.

> if a newly introduced component can avoid unnecessary dependencies,

This proposal does not add any additional dependencies, as explained in the previous comment.

superstar54:
Hi @edan-bainglass , thanks for the comments!

> Sure, but if we can easily address it now by leveraging aiida-core#6303, maybe we should?
> Sure, but it does increase its coupling. We already use abstraction here and there to reduce strong coupling. We should consider this,

Yes, this has already been done in the current implementation. The scheduler does not explicitly talk to RMQ. Instead, it uses the "Abstract message broker" api to communicate with the message broker. The message broker can be RMQ, or other.

